PR #33, opened 2015-12-04
Prototype Java Api

paulcleary:
This is the second revision of the Java / "new" Money API.  The goals are:

- Create a JAVA api so we can support alternative implementations without _requiring_ Scala
- Have an API that does not _assume_ using ThreadLocal for the implementation.  i.e., users can mange their own spans if they like to.

### DIFF
--- a/money-api/src/main/java/com/comcast/money/api/Span.java
+++ b/money-api/src/main/java/com/comcast/money/api/Span.java
@@ -1,0 +1,129 @@
+/*
+ * Copyright 2012-2015 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.comcast.money.api;
+
+import java.util.Map;
+
+/**
+ * A Span is a container that represents a unit of work.  It could be a long running operation or sequence of
+ * statements in process, or a remote system call.
+ *
+ * A Span is immutable, all changes to the span result in a new Span being created.
+ */
+public interface Span {
+
+    /**
+     * Signals the span that it has started
+     */
+    void start();
+
+    /**
+     * Stops the span asserts a successful result
+     */
+    void stop();
+
+    /**
+     * Ends a span, moving it to a Stopped state
+     * @param result The result of the span (success or failure)
+     */
+    void stop(boolean result);
+
+    /**
+     * Records a given note onto the span.  If the note was already present, it will be overwritten
+     * @param note
+     */
+    void record(Note<?> note);
+
+    /**
+     * Starts a new timer on the span
+     * @param timerKey The name of the timer to start
+     * @return a new Span with the timer started
+     */
+    void startTimer(String timerKey);
+
+    /**
+     * Stops an existing timer on the span
+     * @param timerKey The name of the timer
+     * @return a new Span with the timer stopped
+     */
+    void stopTimer(String timerKey);
+
+    /**
+      * @return a map of all of the notes that were recorded on the span.  Implementers should enforce
+     * that the map returned is a copy of the notes
+     */
+    Map<String, Note<?>> notes();
+
+    /**
+     * @return the time in milliseconds when this span was started
+     */
+    Long startTime();
+
+    /**
+     * @return the time in microseconds when this span was started
+     */
+    Long startInstant();
+
+    /**
+     * @return the time in milliseconds when this span was ended.  Will return
+     * null if the span is still open
+     */
+    Long endTime();
+
+    /**
+     * @return the time in microseconds when this span was stopped.
+     */
+    Long endInstant();
+
+    /**
+     * @return the result of the span.  Will return null if the span was never stopped.
+     */
+    Boolean success();
+
+    /**
+     * @return the SpanId of the span.
+     */
+    SpanId id();
+
+    /**
+     * @return the name of the span
+     */
+    String name();
+
+    /**
+     * @return how long since the span was started.  Once it is stopped, the duration should reperesent
+     * how long the span was open for.
+     */
+    Long duration();
+
+    /**
+     * Creates a new child from this span
+     *
+     * @param childName The name of the child to create
+     * @param sticky True if the sticky notes from this span should be passed to the child
+     * @return A new Span that is a child of this span
+     */
+    Span childSpan(String childName, boolean sticky);
+
+    /**
+     * Creates a new child from this span
+     *
+     * @param childName The name of the child to create
+     * @return A new Span that is a child of this span
+     */
+    Span childSpan(String childName);
+}
--- a/money-api/src/main/java/com/comcast/money/api/SpanHandler.java
+++ b/money-api/src/main/java/com/comcast/money/api/SpanHandler.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2012-2015 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.comcast.money.api;
+
+/**
+ * A handler that does some processing on a Span.
+ * <p>
+ *     When a {@link Span} is stopped, the resulting span is passed to the SpanHandler for processing.  Useful
+ *     processing includes things like logging the span, or sending the span to Graphite.
+ * </p>
+ */
+public interface SpanHandler {
+
+    /**
+     * Handles a span that has been stopped.
+     *
+     * @param span {@link Span} that has been stopped and is ready for processing
+     */
+    void handle(Span span);
+}
--- a/money-api/src/main/java/com/comcast/money/api/SpanId.java
+++ b/money-api/src/main/java/com/comcast/money/api/SpanId.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2012-2015 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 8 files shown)


money-api/src/main/java/com/comcast/money/api/Span.java
```diff
@@ -0,0 +1,75 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.comcast.money.api;
+
+/**
+ * A Span is a container that represents a single invocation of a traced service / operation.
```
kristomasette: not necessarily an invocation.

A Span a respressentation of a basic unit of work.  Could be a method invocation, a single operation or a group of operations.  Offers a place to collect Notes describing the span.

money-api/src/main/java/com/comcast/money/api/SpanId.java
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * A unique identifier for a Span
+ */
+public class SpanId {
+
+    private static final Random rand = new Random();
+    private static final String STRING_FORMAT = "SpanId~%s~%s~%s";
+
+    private final String traceId;
+    private final Long parentId;
```
kristomasette: can these be primitive longs?  They shouldn't ever be null right?

money-api/src/test/scala/com/comcast/money/api/DefaultSpanHandler.scala
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.comcast.money.api
+
+class DefaultSpanHandler extends SpanHandler {
+
+  override def handle(spanData: SpanData): Unit = println(spanData.toString)
```
kristomasette: Can we call this LoggingEmittingHandler or maybe Log4JEmittingHandler

I don't think we should just println by default ...

money-api/src/test/scala/com/comcast/money/api/DefaultSpan.scala
```diff
@@ -0,0 +1,99 @@
...
+  private var success: Boolean = true
+
+  override def startTime(): Long = startedTime
+
+  override def start(): Unit = {
+    startedTime = System.currentTimeMillis
+    startInstant = System.nanoTime
+
+    // Start watching me
+    spanMonitor.watch(this)
```
kristomasette: What do you think of keeping the Span management outside of the Span implementation.

The default implementation of Span management should continue to be the akka implementation we have in master, providing us facilities to handle things like backoff and graceful failure.  I see this implementation being best suited for the vast majority of our potential users who will most likely be using Money for instrumentation of Services running in an environment where library footprint is not a concern.

We can bundle a separate implementation in a separate sub-module or project for the few users who are concerned with library footprint or who might be interested in managing span lifecycle on their own.  They would get the more limited Monitor implementation of Span management giving up some resiliency and fault tolerance for a smaller library footprint.

mpilquist: :+1: with keeping span management separate and using existing implementation in master.

money-api/src/main/java/com/comcast/money/api/Span.java
paulcleary: ok

money-api/src/test/scala/com/comcast/money/api/DefaultSpanHandler.scala
paulcleary: This is just a sample implementation inside `src/test` for right now, not meant to actually go into production.

money-api/src/test/scala/com/comcast/money/api/DefaultSpan.scala
paulcleary: See comment above, this is a test implementation not intended to be "real production"; just exercising the API

paulcleary: wrt the existing implementation, I think we will have a `money-akka` or similarly named module that will be the current implementation.  However, I think there is value in having a `money-scala` that doesn't rely on Akka as the implementation.  Akka has historically had version / upgrade issues.

Once we are ok with the API, the next step is to "port" the current implementation as well as get started on a `money-scala` module that can avoid akka and thread local things.  That will allow us to continue to iterate on the 0.9 API.

money-api/src/test/scala/com/comcast/money/api/DefaultSpanHandler.scala
kristomasette: ok..  makes sense

money-api/src/test/scala/com/comcast/money/api/DefaultSpan.scala
kristomasette: Let's discuss more what the default implementation for Span management should be.  Ideally most users should not be burdened with having to weigh the merits of `money-akka` vs. `money-scala` vs. `money-scala` vs `money-java`  

I propose we end up with a single dependency that include the Java API as you defined, A scala implementation of Span and Span Handler, and a scala implementation of Span management that may or not be Akka based depending on whether we think the benefits we get from Akka are worth the potential issues we may run into upgrading Akka versions and/or dealing projects that already include Akka.  That is a discussion we can delay for now as I don't think its a blocker for ironing out the API.  Although it probably makes sense to stick w/ the status quo for now.

There will be another library for those who want to manage Spans on their own.  This would include the java-api as well as the scala implementation of Span and maybe some SpanHandlers.  Or maybe its just the java-api.

And finally there will be a really slim library money-micro (or whatever) that includes the java-api a java-implementation of Span and SpanHandler and a java implementation of SpanManagement without Akka.  That could totally end up being pretty much what you have here in this patch set.

money-api/src/main/java/com/comcast/money/api/SpanHandler.java
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.comcast.money.api;
+
+public interface SpanHandler {
+
+    void handle(Span span);
```
clinedome-work: I'm assuming this is essentially a callback for when a Span is stopped, is that right? Be nice to throw a little documentation here.

money-api/src/main/java/com/comcast/money/api/SpanId.java
clinedome-work: Seems like as-is, `parentId` and `selfId` will be null if we use either of the first two constructors. Which makes sense, I guess, since sometimes there isn't a parent span. Right? Also I'm not 100% sure what a selfId is used for. We should maybe add documentation to make the purpose of each of these clear.

money-api/src/main/java/com/comcast/money/api/Span.java
```diff
@@ -0,0 +1,119 @@
...
+    Long duration();
+
+    /**
+     * Creates a new child from this span
+     *
+     * @param childName The name of the child to create
+     * @param propagate True if the propagated notes from this span should be passed to the child
+     * @return A new Span that is a child of this span
+     */
+    Span childSpan(String childName, boolean propagate);
```
clinedome-work: Ohhh, I had never really noticed the childSpan piece before. This essentially lets you pop open a new namespace for delegated work, yeah? That is a really useful feature.

money-api/src/main/java/com/comcast/money/api/SpanId.java
```diff
@@ -0,0 +1,106 @@
...
+
+        String[] parts = spanIdString.split(",");
+        if (parts.length < 4) {
+            return null;
+        }
+
+        String traceId = parts[1].trim();
+        Long parentId = Long.parseLong(parts[2].trim());
+        Long selfId = Long.parseLong(parts[3].trim());
+        return new SpanId(traceId, parentId, selfId);
```
clinedome-work: If it is possible for the parent/self IDs to be null, then this probably ought to gracefully handle that. Seems like they ought to be able to round-trip.

Would also be fine if the above were primitives and defaulted to `-1` or something. Dunno.

clinedome-work:
I'm really liking the direction of this API. I think it pretty much nails what we need. Also really looking forward to what I'm imagining is the next level up, with a Span implementation that is nicely threadsafe and a handler implementation that dumps the log-ingestion-friendly output to an asynchronous slf4j logger. :)

money-api/src/main/java/com/comcast/money/api/SpanId.java
paulcleary: They should always be there, all 3 values must be present, I can make them primitive

money-api/src/main/java/com/comcast/money/api/SpanHandler.java
paulcleary: Yes, it is a callback for processing stopped spans.  Yes, I will definitely document

money-api/src/main/java/com/comcast/money/api/Span.java
paulcleary: Cool, yea, this is new

money-api/src/main/java/com/comcast/money/api/SpanId.java
paulcleary: I agree, this is kinda roughed in as we explore the api / domain concepts

money-api/src/main/java/com/comcast/money/api/SpanId.java
clinedome-work: Oh yeah, I see, I take back my comment. Didn't realize the constructors chained up like that.

kristomasette:
:+1:

money-api/src/main/java/com/comcast/money/api/SpanId.java
```diff
@@ -0,0 +1,106 @@
...
+
+    private final String traceId;
+    private final long parentId;
+    private final long selfId;
+
+    public SpanId() {
+        this(UUID.randomUUID().toString());
+    }
+
+    public SpanId(String traceId) {
```
kristomasette: maybe we should prevent traceId from being null as well.

money-api/src/main/java/com/comcast/money/api/SpanId.java
```diff
@@ -0,0 +1,106 @@
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        SpanId spanId = (SpanId) o;
+
+        if (parentId != spanId.parentId) return false;
+        if (selfId != spanId.selfId) return false;
+        return traceId.equals(spanId.traceId);
```
kristomasette: as is this will NPE if traceId is null

money-api/src/main/java/com/comcast/money/api/SpanId.java
```diff
@@ -0,0 +1,106 @@
...
+
+    }
+
+    @Override
+    public int hashCode() {
+        int result = traceId.hashCode();
```
kristomasette: NPE here too if traceId is null

money-api/src/main/java/com/comcast/money/api/SpanId.java
clinedome-work: As in, check for null and go with a randomUUID if it is?

money-api/src/main/java/com/comcast/money/api/SpanId.java
paulcleary: I will make sure that trace id is not null

money-api/src/main/java/com/comcast/money/api/SpanId.java
paulcleary: I will make sure that trace id is not null

PR #35, opened 2015-12-10
Add MetricRegistry factory

derjust:
This patchset allows to configure a different way on how
`MetricRegistry` objects are created:

Up to now these were created within money itself with no exposition to
the rest of the system. That makes it impossible to register other
Dropwizzard reporters or exporters on it.

With this patchset it is now possible to provide a custom `MetricRegistry`
factory and do the required initialization of/with the `MetricRegistry`
object.

It is compatible to Java so even Java clients (using the Scala money)
can implement and configure a custom factory.

### DIFF
--- a/money-core/src/main/resources/reference.conf
+++ b/money-core/src/main/resources/reference.conf
@@ -39,7 +39,9 @@ money {
     jvm-exit-on-fatal-error = off
   }
 
-  metricRegistryFactory = "com.comcast.money.metrics.DefaultMetricRegistryFactory"
+  metrics-registry {
+    class-name = "com.comcast.money.metrics.DefaultMetricRegistryFactory"
+  }
 
   emitter {
     emitters = [
--- a/money-core/src/main/resources/reference.conf
+++ b/money-core/src/main/resources/reference.conf
@@ -39,6 +39,8 @@ money {
     jvm-exit-on-fatal-error = off
   }
 
+  metricRegistryFactory = "com.comcast.money.metrics.DefaultMetricRegistryFactory"
+
   emitter {
     emitters = [
       {
--- a/money-core/src/main/scala/com/comcast/money/metrics/MetricRegistryFactory.scala
+++ b/money-core/src/main/scala/com/comcast/money/metrics/MetricRegistryFactory.scala
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2012-2015 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.comcast.money.metrics
+
+import com.typesafe.config.Config
+import akka.actor.{ ActorSystem }
+import com.codahale.metrics.{ MetricRegistry, JmxReporter }
+import scala.util.{ Try, Success, Failure }
+
+/*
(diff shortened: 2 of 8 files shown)


money-core/src/main/scala/com/comcast/money/metrics/MetricRegistryFactory.scala
```diff
@@ -0,0 +1,71 @@
...
+  def metricRegistry(config: Config): MetricRegistry = {
+    Try({
+      // Try to create an instance of the custom factory, configured via 'metricRegistryFactory'
+      val realFactory = Class.forName(config.getString("metricRegistryFactory"))
+        .newInstance.asInstanceOf[MetricRegistryFactory]
+
+      // Ask the custom factory for an MetricRegistry - and pass in our configuration so that an implementation
+      // can add their settings in the application.conf, too.
+      realFactory.metricRegistry(config)
+    }) match {
```
kristomasette: I am thinking since we defined the Default in reference.conf and ensure the DefaultMetricRegistryFactory is in the classpath by including it in the library we should probably log loudly if there is an error.

There should be no way we hit that Fauilure case unless something is really wrong right?

money-core/src/main/scala/com/comcast/money/metrics/MetricRegistryFactory.scala
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * Note: This trait should be kept as simple as possible so that the resulting interface can also be implemented
+ * by a Java client custom factory.
+ */
+object MetricRegistryFactory {
+
+  def metricRegistry(config: Config): MetricRegistry = {
+    Try({
+      // Try to create an instance of the custom factory, configured via 'metricRegistryFactory'
+      val realFactory = Class.forName(config.getString("metricRegistryFactory"))
```
kristomasette: could realFactory be a lazy val in MetriRegistryFactory?

money-core/src/main/resources/reference.conf
```diff
@@ -39,6 +39,8 @@ money {
     jvm-exit-on-fatal-error = off
   }
 
+  metricRegistryFactory = "com.comcast.money.metrics.DefaultMetricRegistryFactory"
```
kristomasette: maybe put this in its own subheading..

metrics-registry {
     class-name = "..."
}

money-core/src/main/scala/com/comcast/money/metrics/MetricRegistryFactory.scala
derjust: Absolutly

money-core/src/main/resources/reference.conf
derjust: Like the idea :)

money-core/src/main/scala/com/comcast/money/metrics/MetricRegistryFactory.scala
derjust: Got your point. 
Followed the pattern from the overall Money initialization: https://github.com/Comcast/money/pull/19/files#r44557079

kristomasette: ok, I guess it makes sense to stop and yell loudly

money-core/src/main/scala/com/comcast/money/metrics/MetricRegistryFactory.scala
```diff
@@ -0,0 +1,74 @@
...
+ * Note: This trait should be kept as simple as possible so that the resulting interface can also be implemented
+ * by a Java client custom factory.
+ */
+object MetricRegistryFactory {
+  private val logger = LoggerFactory.getLogger("com.comcast.money.metrics.MetricRegistryFactory")
+
+  def metricRegistry(config: Config): MetricRegistry = {
+    Try({
+      // Try to create an instance of the custom factory, configured via 'metricRegistryFactory'
+      lazy val realFactory = Class.forName(config.getString("metrics-registry.class-name"))
```
kristomasette: actually  can this lazy val realFactory be at the MetricRegistryFactory level?

That way it gets built once the first time its accessed.

kristomasette:
:+1:

PR #45, opened 2016-01-04
Money core scala

paulcleary:
This is the initial WIP of a *NEW* version of money core.

* Please avoid nit-picking at this time.  I did not write unit tests, integration tests, and commented very little.  I want to get eyes on the design rather than the actual code *

I decided to create a new module for the time being called `money-core-scala`.   Having a separate module will allow us to do A/B testing against the existing money-core, plus we can review this without being distracted from a bunch of change/delete code lines in the current money-core.

### API CHANGES
I wanted to make a few small API changes once I got rolling here:

- Moved creation of child spans to the SpanFactory.  It became evident that there could be some complexity in creating spans, and that responsibility rightly belongs in the factory.
- Separated the span data from the span itself.  There is a new type `SpanInfo` that houses the state/result of a Span.  I did not like sending a `Span` to a `SpanHandler` as that would give developers the opportunity to do stupid things like call `start` and `stop` on a span that was already completed.  Plus, the `SpanInfo` is immutable, which gives additional safety
- Changed the `stop` method to take a `Boolean` (capital B), this simplified implementation in Scala

### NEW CORE
So, the new Core will keep the existing Tracer interface to remain backward compatible.  As such, I mostly copied over the `Tracer` and the `SpanLocal` from the current money-core module.

Of course, the new money-core-scala module uses our new API and the SpanFactory.

The base core is very simple:

- Tracer - creates Spans using the SpanFactory provided and manages them in ThreadLocal (SpanLocal)
- CoreSpan - simple interface; on `stop` will call a `SpanHandler`
- CoreSpanFactory - creates spans and child spans.  You will see that it joins the Span to the SpanHandler so that the CoreSpan `stop` can invoke the SpanHandler

That is the simple part, the complexity is introduced in constructing the Span Handlers based on configuration.  This prototype supports:

- configuring `async`; if async is configured then the handler will be wrapped in an `AsyncSpanHandler`
- setting up a chain of handlers - this is similar to our array of Emitters we have today.  We iterate when we handle spans as opposed to tossing everything into its own future, but that is a small detail that we can change
- HandlerFactory - will create instances of each handler that is configured.  If the handler implements `Configurable`, then that method will be invoked before the handler is done being created

Finally, there is a new `Money` implementation.  No, it is not fully built yet, but I didn't want to add unnecessary complexity until everyone reviewed this implementation.  Any needed things will be moved over.  The new money stuff provides a `Money.Environment`, this will ease testability of the application.  You can see that it supports basic money environment loading, including a disabled environment.

### DIFF
--- a/money-core-scala/src/main/scala/com/comcast/money/core/CoreSpan.scala
+++ b/money-core-scala/src/main/scala/com/comcast/money/core/CoreSpan.scala
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2012-2015 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.comcast.money.core
+
+import java.lang.Long
+
+import com.comcast.money.api._
+
+import scala.collection.JavaConversions._
+import scala.collection.concurrent.TrieMap
+
+/**
+ * A mutable implementation of the Span that also includes SpanInfo
+ *
+ * @param id The [[SpanId]]
+ * @param name The name of the span
+ * @param handler The [[SpanHandler]] responsible for processing the span once it is stopped
+ */
+case class CoreSpan(
+    id: SpanId,
+    name: String,
+    handler: SpanHandler
+) extends Span {
+
+  private var startTimeMillis: Long = 0L
+  private var startTimeMicros: Long = 0L
+  private var endTimeMillis: Long = 0L
+  private var endTimeMicros: Long = 0L
+  private var success: java.lang.Boolean = true
+
+  // use concurrent maps
+  private val timers = new TrieMap[String, Long]()
+  private val noted = new TrieMap[String, Note[_]]()
+
+  def start(): Unit = {
+    startTimeMillis = System.currentTimeMillis
+    startTimeMicros = System.nanoTime / 1000
+  }
+
+  def stop(): Unit = stop(true)
+
+  def stop(result: java.lang.Boolean): Unit = {
+    endTimeMillis = System.currentTimeMillis
+    endTimeMicros = System.nanoTime / 1000
+    this.success = result
+
+    handler.handle(info)
+  }
+
+  def stopTimer(timerKey: String): Unit =
+    timers.remove(timerKey) foreach {
+      timerStartInstant =>
+        record(Note.of(timerKey, System.nanoTime - timerStartInstant))
+    }
+
+  def record(note: Note[_]): Unit = noted += note.name -> note
+
+  def startTimer(timerKey: String): Unit = timers += timerKey -> System.nanoTime
+
+  def info(): SpanInfo =
+    CoreSpanInfo(
+      id,
+      name,
+      startTimeMillis,
+      startTimeMicros,
+      endTimeMillis,
+      endTimeMicros,
+      calculateDuration,
+      success,
+      noted.toMap[String, Note[_]]
+    )
+
+  private def calculateDuration: Long =
+    if (endTimeMicros <= 0L && startTimeMicros <= 0L)
+      0L
+    else if (endTimeMicros <= 0L)
+      (System.nanoTime() / 1000) - startTimeMicros
+    else
+      endTimeMicros - startTimeMicros
+}
--- a/money-core-scala/src/main/scala/com/comcast/money/core/CoreSpanFactory.scala
+++ b/money-core-scala/src/main/scala/com/comcast/money/core/CoreSpanFactory.scala
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2012-2015 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.comcast.money.core
+
+import com.comcast.money.api.{ Span, SpanFactory, SpanHandler, SpanId }
+
+import scala.collection.JavaConversions._
+
+class CoreSpanFactory(handler: SpanHandler) extends SpanFactory {
+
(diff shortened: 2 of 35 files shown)


money-core-scala/src/main/scala/com/comcast/money/core/CoreSpan.scala
```diff
@@ -0,0 +1,85 @@
...
+      id,
+      name,
+      startTime,
+      startInstant,
+      endTime,
+      endInstant,
+      endInstant - startInstant,
+      success,
+      noted.toMap[String, Note[_]]
+    )
```
clinedome-work: So should happen if `.info` is called on a Span that hasn't been stopped? Seems like this is a snapshot of what's currently in the Span, but the duration is going to be all wonky if it hasn't been stopped yet.

money-core-scala/src/main/scala/com/comcast/money/core/CoreSpanFactory.scala
```diff
@@ -0,0 +1,47 @@
...
+  def childSpan(childName: String, span: Span): Span = childSpan(childName, span, false)
+
+  def childSpan(childName: String, span: Span, sticky: Boolean): Span = {
+    val child = newSpan(span.info.id.newChildId, childName)
+
+    if (sticky) {
+      span.info.notes.values
+        .filter(_.isSticky)
+        .foreach(child.record)
+    }
```
clinedome-work: So if "sticky" is true, the new child gets all of its parent's notes? Is this a use case we've seen before?

money-api/src/main/java/com/comcast/money/api/SpanFactory.java
```diff
@@ -20,5 +20,9 @@
 
     Span newSpan(String spanName);
 
-    Span newSpan(String spanName, boolean sticky);
+    Span newSpan(SpanId spanId, String spanName);
+
+    Span childSpan(String childName, Span span);
```
clinedome-work: So if we want to make a child, we send in the parent to the factory? Seems like there might be an easier way... like the parent keeps a reference to the factory that created it (or something). Not a huge deal, just feels a little awkward from an API perspective.

money-core-scala/src/main/scala/com/comcast/money/core/CoreSpan.scala
paulcleary: right, it would definitely be weird, a negative number in that case.

Seems like it would be better to have the duration be the time since the startInstant, if there is no endInstant use System.nanoTime instead, right?

money-core-scala/src/main/scala/com/comcast/money/core/CoreSpanFactory.scala
paulcleary: @clinedome-work yes, we added this, it is an existing feature today.  In CAM and XfinityAPI, Wally wanted to see things like RecordingID and AccountNumber on ALL Spans, so that he can do some interesting queries.  Doing things like Transactions in SPLUNK wound up being a PITA.

So, when we record, we have the option to mark a note as "Sticky", in which case all Stick notes automatically propagate to child spans.

money-api/src/main/java/com/comcast/money/api/SpanFactory.java
paulcleary: @clinedome-work haha, that is the way I had it originally, but I felt like there was a growing sophistication in the way spans were created, that is why I moved the behavior from the span to the span factory.

Putting `childSpan` on the Span itself implied that a Span instance had all of the smarts to create a child, and I was uncomfortable with that.

money-core-scala/src/main/scala/com/comcast/money/core/CoreSpan.scala
clinedome-work: Yeah, that's what I'd think. Could just make them optional in this class and getOrElse them. Or really, that logic could move inside the SpanInfo, though that's nicer just as a dumb bucket.

So why is it we're recording start/stops in both millis and nanos?

paulcleary: duration will ultimately be in microseconds because that is the standard set forth when MBL put together the whole head waters project.  Microseconds is the finiest resolution common across all (most) platforms that we run in

Nanotime does not necessarily coincide with Clock time, System.currentTimeMillis does, so in order to capture the actual real GMT when the thing happened, we need to use System.currentTimeMillis.

clinedome-work: Could maybe then just simplify it to `startTimeMillis, endTimeMillis, durationMicros` and just make the implementing class deal with it. I don't think it would be entirely crazy to see a `SpanInfo` with an `endTimeMillis` of `0L`; that could mean that it just hasn't been stopped yet (or we could add a `stopped` flag, if checking the info of a non-stopped span is something that is actually likely). Would be nice to see as real as possible a duration always, though.

All that said, I'm judging this from the perspective of someone just looking at the API, rather than looking at how it's generally used. I know we hit the `info` function to grab the notes when creating a sticky child span. What other ways is it used, other than in handling a stopped span?

money-core-scala/src/main/scala/com/comcast/money/core/CoreSpanFactory.scala
clinedome-work: Ahhh, gotcha. Is that the "propagate" field on `.record`? May want to get to a common vocabulary on those.

clinedome-work:
Definitely needs more than just my eyes on it, but I really like the looks of this, at least the parts that I'm planning to use. Thanks for all your work on this front. I'll probably have a few nitpicks when it comes time for that, but for now I'm pretty pleased.

money-core-scala/src/main/scala/com/comcast/money/core/CoreSpan.scala
paulcleary: Right, the realistic usage is doing something with a stopped span, we never do anything otherwise (currently).

Your recommendations sound good, I will incorporate those in an update

money-core-scala/src/main/scala/com/comcast/money/core/CoreSpanFactory.scala
paulcleary: Yep, thought the same, there is this `propagate` vs `sticky` thing... will likely update everything to `sticky` in an update

paulcleary: @clinedome-work So, I will use `sticky` in all the places moving forward.

The only place I cannot is in the `TracedData` annotation, it is called "propagate" and I don't want to break backward compatibility with the existing API which would force clients to change code to upgrade to 0.9

clinedome-work: Ok, that's fine. So when you create a child Span here, it only gets the parent's sticky notes if the sticky flag parameter in the `childSpan` invocation is set to true? Seems like if you're recording notes as sticky, you'd expect them to be propagated to all created children, at least by default. Though perhaps that's another case of my not fully understanding the general usage patterns for this feature.

paulcleary: right, you expect sticky notes to propagate to all children

paulcleary:
@kristomasette, *sticky* applies to the note itself.  You are correct, other places should indicate *propagate* or *inheritStickyNotes* or something similar.  I will see if I can work on that

paulcleary:
@kristomasette - you would only have one handler.  We manage multiple handlers in the `HandlerChain`.  This gives us a ton of flexibility in managing our system.

paulcleary:
@kristomasette I did change the names to startTimeMillis and startTimeMicros, same with end, that is in the latest update

PR #59, opened 2016-02-18
Fix adding traceHeader to execute methods with responsehandlers

xfsiva:
Fix adding traceHeader to execute methods with responsehandlers in TraceFriendlyHttpClient


xfsiva:
DId the push in wrong fork

Issue #63, opened 2016-02-27
LogEmitter output is not Splunk friendly


weggert:
There are some aspects of the LogEmitter that are not very friendly for Splunk.

Values for notes are not quoted, which causes Splunk to truncate the extracted value.  For example:
`[ userAgent=Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:44.0) Gecko/20100101 Firefox/44.0 ]`
winds up with this in Splunk:
`userAgent=Mozilla/5.0`

Also the field identifiers for span-name, span-id, trace-id, etc use `-`, which Splunk translates into a `_`.  That means that all searches need to make that translation.

For example, to search for a particular trace-id in Splunk, the query needs to be
`trace_id=<trace-id value>`

I would like to create an improved log format that is more Splunk friendly.  For backwards compatibility I think it may be better not to change the LogEmitter object.  Would creating a SplunkFriendlyLogEmitter object be a reasonable option?

kristomasette:
Sounds totally reasonable to me!

weggert:
Changing the underscores gets a bit more complicated than it would appear at first blush.  The logging keys for that are in money-http-client/src/main/scala/com/comcast/money/http/client/HttpTraceConfig.scala.

Also it would be great if we could have the duration logged in both ms and ns.  Most usages that I have come across are in ms and require a conversion from the provided ns value.

Having something like span_duration=<ms> span_duration_ns=<ns> http_call_duration=<ms> http_call_duration_ns=<ns> would be much easier to work with.

weggert:
After looking at the LogEmitter a bit harder, it looks like we can accomplish everything we need by adding some configuration items to the existing class.

I put a pull request together with the additional configuration.

PR #64, opened 2016-03-01
Add configuration for the LogEmitter format

weggert:
This change updates the LogEmitter to allow the format and keys to be
configured to allow customization of the log output.  See
https://github.com/Comcast/money/issues/63

All of the configuration values have a fallback to the current value,
leaving the current log format unchanged.

The template for logging individual values can be specified as
a log-template property.  The span-start string and null-value can also
be customized.

Each of the field keys can be cutomized with these properties:
  span-id = "span_id"
  trace-id = "trace_id"
  parent-id = "parent_id"
  span-name = "span_name"
  app-name = "app_name"
  start-time = "start_time"
  span-duration = "span_duration_ns"
  span-duration-ms = "span_duration"
  span-success = "span_success"

Also a new span-duration-ms field is added to allow logging the duration
in ms.  By default it is disabled, however the span-duration-ms-enabled
configuration value can be used to enable it.

### DIFF
--- a/money-core/src/main/resources/reference.conf
+++ b/money-core/src/main/resources/reference.conf
@@ -54,6 +54,23 @@ money {
         configuration = {
           emitter = "com.comcast.money.emitters.LogEmitter"
           log-level = "WARN"
+          formatting {
+            span-start = "Span: "
+            null-value = "NULL"
+            log-template = "[ %s=%s ]"
+            span-duration-ms-enabled = "false"
+            keys {
+              span-id = "span-id"
+              trace-id = "trace-id"
+              parent-id = "parent-id"
+              span-name = "span-name"
+              app-name = "app-name"
+              start-time = "start-time"
+              span-duration = "span-duration"
+              span-duration-ms = "span-duration-ms"
+              span-success = "span-success"
+            }
+          }
         }
       },
       {
--- a/money-core/src/main/scala/com/comcast/money/emitters/LogEmitter.scala
+++ b/money-core/src/main/scala/com/comcast/money/emitters/LogEmitter.scala
@@ -27,9 +27,6 @@ import scala.util.Try
 
 object LogEmitter {
 
-  val logTemplate = "[ %s=%s ]"
-  val NULL = "NULL"
-
   def props(conf: Config): Props =
     Try {
       val emitter = conf.getString("emitter")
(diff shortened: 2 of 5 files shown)


money-core/src/main/scala/com/comcast/money/emitters/LogEmitter.scala
```diff
@@ -38,34 +35,48 @@ object LogEmitter {
...
-    builder.append("Span: ")
-    append("span-id", t.spanId.spanId)
-    append("trace-id", t.spanId.traceId)
-    append("parent-id", t.spanId.parentSpanId)
-    append("span-name", t.spanName)
-    append("app-name", Money.applicationName)
-    append("start-time", t.startTime)
-    append("span-duration", t.duration)
-    append("span-success", t.success)
+    builder.append(configValue("span-start", "Span: "))
```
paulcleary: A few comments on this...

It would seem to make sense to parse the Config into some type and use that as a val first.  

Rather than having a method on LogEmitter that gets pulled in, just have a class.  Some psuedo code:

```scala
object LogFormatter {

  // by providing a default setup, you don't have to check has path every time you log
  val default =
    """
      |keys {
      |   span-id = "span-id"
      |   trace-id = "trace-id"
      |   parent-id = "parent-id"
      |   span-name = "span-name"
      |   app-name = "app-name"
      |   start-time = "start-time"
      |   span-duration = "span-duration"
      |   span-success = "span-success"
      |}
    """.stripMargin

  val defaultConfig = ConfigFactory.parseString(default)

  def apply(conf: Config): LogFormatter = {
    // the passed in config with fallback will ensure there are default values
    val config = conf.withFallback(defaultConfig)

    // build the log formatter with the config values.  You can create this once when the actor is created...
    new LogFormatter(config.getString("keys.span-id"), config.)
  }
}

/* By having the loging done here, you don't have to inspect the config when logging */
class LogFormatter(val spanId: String, val traceId: String, val parentId: String, val spanName: String,
                   val appName: String, val startTime: String, val spanDuration: String, val spanSuccess: String) {

  def format(span: Span): String = ??? // TODO: fun-times here
}
```

paulcleary: My concern is having to do all that checking every time we build a message.  Would be better if the config was parsed out to values and usable.

The config only needs to be inspected when the system starts up.

The above is one way to do it, I am sure there are other ways.

weggert: I agree 100% with your comment.  I have pushed up another commit that pulls out the property parsing into a separate utility class and does the parsing only once.

paulcleary:
This is better.

My response is that it would be great if the "Logger" was a separate class all together.

The reason is self-serving, as I can then re-use your code in 0.9 with very little if any modification :)

Let me know if that's asking too much.  I am just thinking a POJO type class that actually does the logging.

You have a case class `LogEmitterProperties` here.

If you made it something like a `SpanLogFormatter` with the same properties, then we could move all the string generation to the SpanLogFormatter.  It would make it easier to test and simpler to move to 0.9

Anyway, this is a conversation, let me know your thoughts.

All the Actor does is:

```scala
val formatter = SpanLogFormatter(config)
```

money-core/src/main/resources/reference.conf
```diff
@@ -54,6 +54,21 @@ money {
         configuration = {
           emitter = "com.comcast.money.emitters.LogEmitter"
           log-level = "WARN"
+          span-start = "Span: "
+          null-value = "NULL"
```
kristomasette: does it make sense for these to be under the emitter configuration?

maybe something like:
emitter = {
     class="com.comcast.money.emitters.LogEmitter
     formatting = {
          span-start = "Span: "
          ...
    }
}

money-core/src/main/scala/com/comcast/money/emitters/LogEmitter.scala
```diff
@@ -38,51 +91,47 @@ object LogEmitter {
       case _ => Props(classOf[LogEmitter], conf)
     }.get
 
-  def buildMessage(t: Span): String = {
+  def buildMessage(t: Span)(implicit emitterProps: LogEmitterProperties): String = {
```
kristomasette: why is emitterProps implicit?

weggert:
Separating out the LogFormatter from the LogEmitter worked out pretty well.  It makes for a better separation of concerns for the classes.

The latest patch includes the separation and also moves the properties into a 'formatting' group per Kris Tomasette's suggestion.

money-core/src/test/scala/com/comcast/money/emitters/LogFormatterSpec.scala
```diff
@@ -0,0 +1,156 @@
...
+        """
+      )
+      val spanLogFormatter = SpanLogFormatter(conf)
+      val sampleData = Span(
+        SpanId(1L), "key", "unknown", "host", 1L, true, 35L,
+        Map("what" -> Note("what", 1L), "when" -> Note("when", 2L), "bob" -> Note("bob", "craig"))
+      )
+      val actualMessage = spanLogFormatter.buildMessage(sampleData)
+      assert(
+        actualMessage === ("Span: span-id=\"1\" trace-id=\"1\" parent-id=\"1\" span-name=\"key\" " +
```
paulcleary: Instead of escaping the quotes here, you could use the triple quote I think, then ditch those escape characters.

```scala
actualMessage === """
```

paulcleary:
This is great, one small suggestion there about the escape characters and then :+1: from me.

kristomasette:
:+1:

PR #65, opened 2016-03-14
Add money tracing support to AmazonWebServiceClient.

ipapa:
- AmazonWebServiceClient is what most aws-java-sdk platform services use internally.
DynamoDB client is a good example that makes use of this client.
- AmazonWebServiceClient encapsulates the actual instance of the HttpClient and does not
want its clients to interact with it directly. The only way you can do so is through RequestHandlers.
- This implements a request handler to AmazonWebServiceClient that sends the money trace data
to the platform as a HTTP request header. Once the platform service reads the trace data from
the client request object, it is supposed to span on that.
- This was tested just with the AmazonDynamoDBClient but it should work with any of the other
 AWS platform services that make use of the AmazonWebServiceClient internally.

### DIFF
--- a/money-aws-java-sdk/src/main/scala/com/comcast/money/java/AmazonWebServiceClientMoneyTraceRequestHandler.java
+++ b/money-aws-java-sdk/src/main/scala/com/comcast/money/java/AmazonWebServiceClientMoneyTraceRequestHandler.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2012-2015 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.comcast.money.java;
+
+import com.amazonaws.Request;
+import com.amazonaws.Response;
+import com.amazonaws.handlers.RequestHandler2;
+import com.comcast.money.core.SpanId;
+import com.comcast.money.internal.SpanLocal;
+import com.comcast.money.japi.JMoney;
+
+import scala.Option;
+
+public class AmazonWebServiceClientMoneyTraceRequestHandler extends RequestHandler2 {
+
+    public static final String MONEY_TRACE_HEADER = "X-MoneyTrace";
+
+    private final String spanName;
+
+    public AmazonWebServiceClientMoneyTraceRequestHandler(String spanName) {
+
+        this.spanName = spanName;
+    }
+
+    @Override
+    public void beforeRequest(Request<?> request) {
+
+        super.beforeRequest(request);
+
+        if (!JMoney.isEnabled()) {
+            return;
+        }
+
+        JMoney.startSpan(spanName);
+        Option<SpanId> spanId = SpanLocal.current();
+        if (spanId.isDefined()) {
+            request.addHeader(MONEY_TRACE_HEADER, spanId.get().toHttpHeader());
+        }
+    }
+
+    @Override
+    public void afterResponse(Request<?> request, Response<?> response) {
+
+        super.afterResponse(request, response);
+
+        if (!JMoney.isEnabled()) {
+            return;
+        }
+
+        JMoney.stopSpan(true);
+    }
+
+    @Override
+    public void afterError(Request<?> request, Response<?> response, Exception e) {
+
+        super.afterError(request, response, e);
+
+        if (!JMoney.isEnabled()) {
+            return;
+        }
+
+        JMoney.stopSpan(false);
+    }
+}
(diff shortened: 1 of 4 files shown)


money-aws-java-sdk/src/main/scala/com/comcast/money/java/AmazonWebServiceClientMoneyTraceRequestHandler.java
```diff
@@ -0,0 +1,78 @@
...
+            request.addHeader(MONEY_TRACE_HEADER, spanId.get().toHttpHeader());
+        }
+    }
+
+    @Override
+    public void afterResponse(Request<?> request, Response<?> response) {
+
+        super.afterResponse(request, response);
+
+        if (!JMoney.isEnabled()) {
```
pcleary00: would it make sense to have the super.afterReponse go inside a `try` and all the money stuff go inside a `finally`?  Not sure if `super.afterResponse` can throw or not, but we always want to do everything we can to ensure we stop spans (lest we will start a memory leak potentially)

money-aws-java-sdk/src/test/scala/com/comcast/money/java/AmazonWebServiceClientMoneyTraceRequestHandlerTest.java
```diff
@@ -0,0 +1,204 @@
...
+    @Test
+    public void testAmazonWebServiceClientMoneyTraceRequestHandler() {
+
+        assertNotNull(handler);
+    }
+
+    @Test
+    public void testBeforeRequestRequestOfQ_moneyNotEnabled() {
+
+        PowerMockito.mockStatic(JMoney.class, SpanLocal.class);
```
pcleary00: I don't like introducing Power Mockito, only because we try and limit the number of things we do / ways we do them in order to make the entire code somewhat consistent.

The only other way to do though would be to wrap the JMoney and SpanLocal calls inside a `MoneyHelper` class that can be mocked.  Thoughts?

money-aws-java-sdk/src/main/scala/com/comcast/money/java/AmazonWebServiceClientMoneyTraceRequestHandler.java
ipapa: I do not think those methods do anything. We can even remove them completely. I added them since I am extending their abstract class.

http://grepcode.com/file_/repo1.maven.org/maven2/com.amazonaws/aws-java-sdk/1.8.1/com/amazonaws/handlers/RequestHandler2.java/?v=source

money-aws-java-sdk/src/test/scala/com/comcast/money/java/AmazonWebServiceClientMoneyTraceRequestHandlerTest.java
ipapa: Not sure how MoneyHelper wrapper will help since the methods we are trying to mock are scope=static in both those classes. I am not a big fan of powermock as well, it is a weapon of last resort to mock static method calls or final classes.

money-aws-java-sdk/src/main/scala/com/comcast/money/java/AmazonWebServiceClientMoneyTraceRequestHandler.java
paulcleary: You do start a span, and it is really important that a span is stopped.

Kinda makes sense in those methods.

I will have to look at the code a little closer.

money-aws-java-sdk/src/test/scala/com/comcast/money/java/AmazonWebServiceClientMoneyTraceRequestHandlerTest.java
paulcleary: What I meant was creating a `MoneyHelper` that is a normal class (or interface), and exposing methods on it like startSpan, stopSpan, and, currentSpanId.  If you set it as a protected member of the class, you can override it in a unit test to be a mock.

money-aws-java-sdk/src/main/scala/com/comcast/money/java/AmazonWebServiceClientMoneyTraceRequestHandler.java
ipapa: Keep in mind that we are in the context of an aws webservice client request handler. The aws sdk is handling errors with a try/catch block so we do not have to here. These three methods are just hooks to before request is issued, to do something, in our case add a header. And we will get a callback when the operation is done: when the HTTP call was a success, afterResponse() will be called, and when the call failed with an exception, afterError() will be called.

Bottom line, we are not wrapping the AmazonWebserviceClient here like we do in the other instances in money. We are just registering this handler as a hook to their workflow.

money-aws-java-sdk/src/test/scala/com/comcast/money/java/AmazonWebServiceClientMoneyTraceRequestHandlerTest.java
ipapa: That change goes beyond the scope of this patch set and it will require some refactoring to the existing library to make it work, it is not as simple as rewriting the test cases. Is it possible to merge this as-is and work on that as a separate patch set?

ipapa:
Also, any idea why the CI build failed? It looks from the logs that one test case failed but I do not think it is related to my changes.

money-aws-java-sdk/src/test/scala/com/comcast/money/java/AmazonWebServiceClientMoneyTraceRequestHandlerTest.java
paulcleary: I suppose I shouldn't recommend the how here.  I really do appreciate the contribution, but given the project is a multi-module project I have to maintain consistency as much as I can.

There are ways to implement mocking without power mockito, for example via sub-classing or via a dependency injection, among others.

If you can update your PR and remove power mockito, we should be good.

money-aws-java-sdk/src/main/scala/com/comcast/money/java/AmazonWebServiceClientMoneyTraceRequestHandler.java
paulcleary: Good point, do we need to call super.beforeRequest and super.afterResponse there since the parent isn't doing anything?

paulcleary:
:+1:

PR #67, opened 2016-04-28
Fix Javadoc issue on compile for publishLocal

paulcleary:
Had to remove straggling `</p>` and other javadoc errors.

Also, slightly changed the build to set javacoptions in `(Compile, doc)` as opposed to just `doc`

This seems to fix the issues with doing a publishLocal

### DIFF
--- a/money-api/src/main/java/com/comcast/money/annotations/Timed.java
+++ b/money-api/src/main/java/com/comcast/money/annotations/Timed.java
@@ -23,14 +23,14 @@ import java.lang.annotation.Target;
 
 /**
  * Used to indicate that the code path should be traced and executes
- * <p/>
+ * <p>
  * This annotation can be used on any method that you want to trace.  Typically this will be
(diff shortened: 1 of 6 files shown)


money-api/src/main/java/com/comcast/money/annotations/Timed.java
```diff
@@ -23,14 +23,14 @@
 
 /**
  * Used to indicate that the code path should be traced and executes
- * <p/>
+ * <p>
  * This annotation can be used on any method that you want to trace.  Typically this will be
  * used on the controller to indicate a new request trace, as well as external service methods
- * <p/>
```
clinedome-work: Shouldn't we at least close the <p> tags we open? I guess it doesn't really matter, but seems like we might as well bring what little balance to the universe we can

clinedome-work:
Worked for me. I don't feel crazy strongly about closing the tags. I have a mild preference towards closing them, but it is up to you. :+1:

PR #68, opened 2016-05-02
Add toString method to Note

clinedome-work:
Was doing some debugging earlier. Most of the other pieces had nice
toString methods, but I couldn't see inside the Note. Quick fix to make
that easier.

Figure we don't test toString methods, though I could add that easily
enough if you wanted.

### DIFF
--- a/money-api/src/main/java/com/comcast/money/api/Note.java
+++ b/money-api/src/main/java/com/comcast/money/api/Note.java
@@ -273,4 +273,16 @@ public class Note<T> {
         result = 31 * result + (sticky ? 1 : 0);
         return result;
     }
+
+    @Override
+    public String toString() {
+        StringBuilder builder = new StringBuilder();
+        builder.append("Note(");
+        builder.append("name=").append(name);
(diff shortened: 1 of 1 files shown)


money-api/src/main/java/com/comcast/money/api/Note.java
```diff
@@ -273,4 +273,16 @@ public int hashCode() {
         result = 31 * result + (sticky ? 1 : 0);
         return result;
     }
+
+    @Override
+    public String toString() {
+        StringBuilder builder = new StringBuilder();
+        builder.append("Note(");
+        builder.append("name=").append(name);
+        builder.append(", value=").append(value.toString());
```
paulcleary: can a value be null?

clinedome-work: Good call. Checking it out, will push up necessary changes.

paulcleary:
Can you add a simple unit test; one that does a normal and one that does the null?

paulcleary:
👍

PR #80, opened 2017-05-04
adds new factory method for creating a child span from an x-moneytrace header

benschmaus:
Adds a new method `childOrRootSpan` on `SpanFactory` that makes it easy to create a child span from a given `x-moneytrace` context header.

If the given `x-moneytrace` header is well formed then a new child span is created with the trace ID and parent ID from the header.

If the `x-moneytrace` header is malformed then a new root span is created.

This removes the need for client code to parse the `x-moneytrace` header and ensure a child span is created using the trace and parent IDs from the header.

### DIFF
--- a/money-api/src/main/java/com/comcast/money/api/SpanFactory.java
+++ b/money-api/src/main/java/com/comcast/money/api/SpanFactory.java
@@ -31,7 +31,7 @@ public interface SpanFactory {
      * @return a child span with trace id and parent id from trace context header or a new root span if the
      * traceContextHeader is malformed.
      */
-    Span childOrRootSpan(String childName, String traceContextHeader);
+    Span newSpanFromHeader(String childName, String traceContextHeader);
 
     Span childSpan(String childName, Span span);
 
--- a/money-api/src/main/java/com/comcast/money/api/SpanFactory.java
+++ b/money-api/src/main/java/com/comcast/money/api/SpanFactory.java
@@ -22,6 +22,17 @@ public interface SpanFactory {
 
     Span newSpan(SpanId spanId, String spanName);
 
+    /**
+     * Continues a trace by creating a child span from the given x-moneytrace header
+     * value.
+     *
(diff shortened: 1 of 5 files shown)


money-api/src/main/java/com/comcast/money/api/SpanFactory.java
```diff
@@ -22,6 +22,17 @@
...
+    /**
+     * Continues a trace by creating a child span from the given x-moneytrace header
+     * value.
+     *
+     * @param childName - the name of the child span to create
+     * @param traceContextHeader - value of x-moneytrace header
+     * @return a child span with trace id and parent id from trace context header or a new root span if the
+     * traceContextHeader is malformed.
+     */
+    Span childOrRootSpan(String childName, String traceContextHeader);
```
paulcleary: Don't like "childOrRootSpan", I think "newSpanFromHeader"

money-core/src/main/scala/com/comcast/money/core/CoreSpanFactory.scala
```diff
@@ -18,12 +18,29 @@ package com.comcast.money.core
...
 
   def newSpan(spanName: String): Span = newSpan(new SpanId(), spanName)
 
+  /**
+   * Continues a trace by creating a child span from the given x-moneytrace header
+   * value or a root span if header is malformed.
+   *
+   * @param childName - the name of the child span to create
+   * @param traceContextHeader - value of x-moneytrace header
+   * @return a child span with trace id and parent id from trace context header or a new root span if the
```
paulcleary: this is interesting, is this what we want to do?

I would think if the traceContextHeader is present and malformed, that would be an exception condition.

Perhaps at the minimum we log the error here.

benschmaus: added a log warning in the case of a malformed `x-moneytrace` header value.

PR #81, opened 2017-08-29
Correcting typo in variable name

petercline:
I noticed com.comcast.money.java.servlet.TraceFilter made repeated
reference to an incTrcaceId.  The pedant and copy editor trapped inside
this programmer's body could not let this alone.

This change is solely to correct this typo and restore some sense of
inner peace to myself and any other similarly afflicted folks who may
happen upon this source code.

### DIFF
--- a/money-java-servlet/src/main/scala/com/comcast/money/java/servlet/TraceFilter.scala
+++ b/money-java-servlet/src/main/scala/com/comcast/money/java/servlet/TraceFilter.scala
@@ -44,13 +44,13 @@ class TraceFilter extends Filter {
 
     SpanLocal.clear()
     val httpRequest = new HttpServletRequestWrapper(request.asInstanceOf[HttpServletRequest])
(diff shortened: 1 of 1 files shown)


petercline:
When I check the cla assistant it tells me the cla is already signed.

petercline:
I've tried signing it multiple times.  Every time I try I'm told it is
already signed.  This has been the case since day one.  I've no idea how to
proceed.

petercline:
The check above says the cla isn't signed.  Every time I try to sign it I'm told it is already signed.  Ths has been the case since I opened the pull request in August.

petercline:
Here's what I see when I try to sign again
screen shot 2018-03-13 at 10 58 44 pm

ptravers:
I'm a bit late to this but I had a similar problem because there was code in the repository written on my work github account so I had to sign in as that user then sign the CLA.

pauljamescleary:
Yes, @petercline The commit says "omicron committed on Aug 29, 2017"

I wonder if that is where it is getting confused.  The commit author is different than "petercline"

PR #84, opened 2018-03-01
Add tracing support for Apache HttpAsyncClient

HaloFour:
Adds trace-friendly helper wrapper around Apache `HttpAsyncClient`.  This is an NIO-based HTTP client which executes the requests asynchronously and uses callback interfaces in order to notify of completion.  The helper class wraps said callback interfaces in order to intercept the HttpRequest and HttpResponse objects in order to instrument them as well as to restore the parent tracing span on the callback thread.

I've not yet looked into AspectJ instrumentation of `HttpAsyncClient`.

This PR also includes a general-purpose mechanism for capturing the current tracing state that can then be restored on a separate thread.  This is following the pattern established in `TraceFriendlyThreadPoolExecutor`.


pauljamescleary:
Thanks for this!  Will review this week.  Looks like travis is unhappy, so should fix that first.

HaloFour:
When working on the unit tests for the aspect intercepting `HttpAsyncClient` calls I noticed something about how spans work in asynchronous contexts.  I had followed the design established in `TraceFriendlyThreadPoolExecutor` where the current span and MDC are captured and then restored onto the worker thread on which the execution will continue.  However, this design assumes that the span submitting the request will remain open for the duration.  When instrumented via `@Traced` annotations the span is closed when the method returns.  That might work fine in a strictly fork-join model, but I envision using this in a reactive model, where the method coordinating the HTTP request returns almost immediately with a `CompletionStage<T>`/`Completable<T>`/`Mono<T>`/`Observable<T>` representing the result.

On the surface it would seem that the appropriate route here would be to have the instrumentation automatically create a child span of the current span.  Per @kristomasette the child span can continue to trace after the parent span is closed and that you can correlate the children back to their parent.  This seems like it will work, although I'll have to think of a good way to derive a name for the child span.

I'd like to mull over the nature of spans and tracing in such an asynchronous context, though, to see if a different approach might fit better.  For the sake of comparison, with the NewRelic API in the context of a transaction (what they call a span) you can obtain a reference of a "token" to that transaction.  You can then link any method on any thread back to that transaction.  The transaction would maintain a count of these tokens and would only consider the transaction complete when both it was closed and all of the tokens have been expired.

The reason I'd like to explorer this further is because I'd like to look into adding more modules for adding instrumentation around common libraries like `RxJava` and Spring Reactor.  Kris has also mentioned wanting to explore better tracing across Akka-actor, perhaps correlating activity in a similar manner to the `X-MoneyTrace` HTTP header.

money-http-client/src/main/scala/com/comcast/money/http/client/TraceFriendlyHttpClient.scala
```diff
@@ -122,12 +122,12 @@ class TraceFriendlyHttpClient(wrapee: HttpClient) extends HttpClient with java.i
   }
 
   override def close(): Unit = {
-    if (wrapee.isInstanceOf[CloseableHttpClient])
-      wrapee.asInstanceOf[CloseableHttpClient].close()
-    else if (wrapee.isInstanceOf[Closeable])
-      wrapee.asInstanceOf[Closeable].close()
-    else if (wrapee.isInstanceOf[AutoCloseable])
-      wrapee.asInstanceOf[AutoCloseable].close()
+    wrapee match {
```
pauljamescleary: nice cleanup.  I think money in general needs a refresh / update because there is likely lots of opportunity for code cleanup

money-http-async-client/src/main/scala/com/comcast/money/http/client/TraceFriendlyHttpAsyncClient.scala
```diff
@@ -0,0 +1,262 @@
...
+    wrappee.failed(ex)
+  }
+
+  override def consumeContent(decoder: ContentDecoder, ioctrl: IOControl): Unit = wrappee.consumeContent(decoder, ioctrl)
+  override def isDone: Boolean = wrappee.isDone
+  override def getResult: T = wrappee.getResult
+  override def getException: Exception = wrappee.getException
+  override def cancel(): Boolean = wrappee.cancel()
+  override def close(): Unit = wrappee.close()
+}
```
pauljamescleary: add a new line to close this file

money-http-async-client/src/main/scala/com/comcast/money/http/client/TraceFriendlyHttpAsyncClient.scala
```diff
@@ -0,0 +1,262 @@
...
+/**
+ * Provides a thin wrapper around [[HttpAsyncClient]] to support automatically tracing
+ * requests and restoring the tracing state on the callback interfaces.
+ */
+class TraceFriendlyHttpAsyncClient(wrappee: HttpAsyncClient) extends HttpAsyncClient
+    with java.io.Closeable {
+
+  import com.comcast.money.http.client.TraceFriendlyHttpAsyncSupport._
+
+  val tracer = Money.Environment.tracer
```
pauljamescleary: would it make sense to pass in the tracer in the constructor?

HaloFour: I was following the convention already set by [`TraceFriendlyHttpClient`](https://github.com/Comcast/money/blob/master/money-http-client/src/main/scala/com/comcast/money/http/client/TraceFriendlyHttpClient.scala#L75).  I don't mind breaking from that convention if it makes more sense.  Perhaps overloaded constructors for both?  Might make mocking easier rather than having to extend from the class which is what the unit tests currently do.

pauljamescleary: Sorry about what is likely some shaky precedent.  I appreciate it that you tried to maintain consistency.  But yea, would make things easier just passing it in.

pauljamescleary: Let me know if you can make that adjustment or if you need me to pair with you on this one.

pauljamescleary: Got it

pauljamescleary:
@HaloFour circling back to this.  I am readying a 0.9.0 release.  Is this something that is still needed?

HaloFour:
Sorry, this slipped off of my radar.  Our team is currently in the process of moving to Java 11 on OpenJDK and my interest has turned to supporting the now built-in `java.net.http.HttpClient` class, which supports both sync and async operations.  Do you think that this would PR would be valuable to others?

Speaking of JDK11, do you know if it's possible to support a project under money that supports the newer runtimes?  Or would that involve updating all of the projects?

pauljamescleary:
@HaloFour no worries on this PR, I was just asking if it was something that you guys specifically used.  If not, we can add it after the 0.9.0 release in a dot release afterwards.

As far as JDK 11, I have updated all of our dependencies.  This is rather significant:

- removed support for Java 7
- Java 8 is the default target environment now for 0.9.x+
- updated all dependencies to be current
- will likely remove support for Spring 3 as well, since that has been EOL for almost 2 years now.
- in theory, we should be fine running in JDK 11 going forward.  Should really update travis to do a dual build on jdk 11

ptravers:
To support JDK 11 we might have to drop Scala 11 support?

PR #86, opened 2018-03-09
Add support for tracing methods asynchronously

HaloFour:
Adds support for tracing methods that return objects indicating asynchronous completion of a given task.  The span itself remains open until the task is completed as determined by an implementation of `AsyncTraceService` that can support the return type of the method and register for completion.  The duration of the method execution itself is logged via a timer on the span.

This PR includes support for Scala's `Future[T]` trait.

### DIFF
--- a/money-aspectj/src/main/scala/com/comcast/money/aspectj/TraceAspect.scala
+++ b/money-aspectj/src/main/scala/com/comcast/money/aspectj/TraceAspect.scala
@@ -18,12 +18,14 @@ package com.comcast.money.aspectj
 
 import com.comcast.money.annotations.{ Timed, Traced }
 import com.comcast.money.core._
-import com.comcast.money.core.internal.MDCSupport
+import com.comcast.money.core.async.AsyncTracingService
+import com.comcast.money.core.internal.{ MDCSupport, SpanLocal }
 import com.comcast.money.core.logging.TraceLogging
 import com.comcast.money.core.reflect.Reflections
 import org.aspectj.lang.annotation.{ Around, Aspect, Pointcut }
 import org.aspectj.lang.reflect.MethodSignature
 import org.aspectj.lang.{ JoinPoint, ProceedingJoinPoint }
+import org.slf4j.MDC
 
 @Aspect
 class TraceAspect extends Reflections with TraceLogging {
@@ -41,12 +43,42 @@ class TraceAspect extends Reflections with TraceLogging {
   def adviseMethodsWithTracing(joinPoint: ProceedingJoinPoint, traceAnnotation: Traced): AnyRef = {
     val key: String = traceAnnotation.value
     var result = true
+    var stopSpan = true
     val oldSpanName = mdcSupport.getSpanNameMDC
+
     try {
       tracer.startSpan(key)
       mdcSupport.setSpanNameMDC(Some(key))
       traceMethodArguments(joinPoint)
-      joinPoint.proceed
+
+      if (traceAnnotation.async()) {
+        val methodTimeKey = "method-time"
+        tracer.startTimer(methodTimeKey)
+        var retval = joinPoint.proceed
+        tracer.stopTimer(methodTimeKey)
+
+        AsyncTracingService.findTracingService(retval).foreach {
+          service =>
+            stopSpan = false
+            val asyncSpan = SpanLocal.current.get
+            val mdc = Option(MDC.getCopyOfContextMap)
+
+            retval = service.whenDone(retval, (_, exception) => {
+              mdcSupport.propogateMDC(mdc)
+              result = true
+              if (exception != null) {
+                logException(exception)
+                result = exceptionMatches(exception, traceAnnotation.ignoredExceptions())
+              }
+              asyncSpan.stop(result)
+              MDC.clear()
+            })
+        }
+
+        retval
+      } else {
+        joinPoint.proceed
+      }
     } catch {
       case t: Throwable =>
         result = if (exceptionMatches(t, traceAnnotation.ignoredExceptions())) true else false
@@ -54,7 +86,11 @@ class TraceAspect extends Reflections with TraceLogging {
         throw t
     } finally {
       mdcSupport.setSpanNameMDC(oldSpanName)
-      tracer.stopSpan(result)
+      if (stopSpan) {
+        tracer.stopSpan(result)
+      } else {
+        SpanLocal.pop()
+      }
     }
   }
 
--- a/money-aspectj/src/test/scala/com/comcast/money/aspectj/TraceAspectSpec.scala
+++ b/money-aspectj/src/test/scala/com/comcast/money/aspectj/TraceAspectSpec.scala
@@ -18,12 +18,15 @@ package com.comcast.money.aspectj
 
 import com.comcast.money.annotations.{ Timed, Traced, TracedData }
 import com.comcast.money.core.internal.MDCSupport
-import com.comcast.money.core.{ Tracer, Money, SpecHelpers }
+import com.comcast.money.core.{ LogRecord, Money, SpecHelpers }
 import org.aspectj.lang.ProceedingJoinPoint
 import org.mockito.Mockito._
 import org.scalatest._
 import org.scalatest.concurrent.Eventually
 import org.scalatest.mock.MockitoSugar
+
+import scala.concurrent.Promise
+import scala.util.Try
 
 class TraceAspectSpec extends WordSpec
     with GivenWhenThen with OneInstancePerTest with BeforeAndAfterEach with Matchers with MockitoSugar with Eventually
@@ -75,6 +78,12 @@ class TraceAspectSpec extends WordSpec
   def methodWithNonMatchingIgnoredException() = {
     throw new RuntimeException("not-ignored")
   }
+
+  @Traced(
+    value = "methodReturnsScalaPromise",
+    async = true
+  )
+  def methodReturnsScalaPromise(promise: Promise[String]): Promise[String] = promise
 
   @Timed("methodWithTiming")
   def methodWithTiming() = {
@@ -144,6 +153,19 @@ class TraceAspectSpec extends WordSpec
 
         And("a span-success is logged with a value of false")
         expectLogMessageContaining("span-success=false")
+      }
+      "handle async methods" in {
+        Given("a method that returns a future")
+
+        When("the method is invoked")
+        val promise = Promise[String]()
+        methodReturnsScalaPromise(promise)
+
+        promise.complete(Try("Success"))
+        expectLogMessageContaining("methodReturnsScalaPromise")
+        expectLogMessageContaining("method-time")
+
+        LogRecord.log("log").foreach(System.out.println(_))
       }
     }
     "advising methods that have parameters with the TracedData annotation" should {
--- a/money-core/src/main/scala/com/comcast/money/core/async/AsyncTracingService.scala
+++ b/money-core/src/main/scala/com/comcast/money/core/async/AsyncTracingService.scala
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2012-2015 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.comcast.money.core.async
+
+import java.util.ServiceLoader
+
+import scala.collection.JavaConverters._
+
+object AsyncTracingService {
+  private lazy val services = loadServices()
+
+  def findTracingService(future: AnyRef): Option[AsyncTracingService] =
+    services.find(_.supports(future))
+
+  private def loadServices() =
+    ServiceLoader.load(classOf[AsyncTracingService], classOf[AsyncTracingService].getClassLoader).asScala.toList
+}
+
+trait AsyncTracingService {
+  def supports(future: AnyRef): Boolean
+  def whenDone(future: AnyRef, f: (Any, Throwable) => Unit): AnyRef
+}
--- a/money-api/src/main/java/com/comcast/money/annotations/Traced.java
+++ b/money-api/src/main/java/com/comcast/money/annotations/Traced.java
@@ -75,4 +75,11 @@ public @interface Traced {
      * @return The array of exception classes that will be ignored
      */
     Class[] ignoredExceptions() default {};
+
+    /**
+     * Indicates that the method being traced performs an asynchronous task and returns an instance
+     * that would indicate when the task has completed.
+     * @return Whether or not the traced method is asynchronous
+     */
+    boolean async() default false;
 }
--- a/money-core/src/main/resources/META-INF/services/com.comcast.money.core.async.AsyncTracingService
+++ b/money-core/src/main/resources/META-INF/services/com.comcast.money.core.async.AsyncTracingService
@@ -1,0 +1,1 @@
+com.comcast.money.core.async.ScalaFutureTracingService
--- a/money-core/src/main/scala/com/comcast/money/core/async/ScalaFutureTracingService.scala
+++ b/money-core/src/main/scala/com/comcast/money/core/async/ScalaFutureTracingService.scala
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2012-2015 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.comcast.money.core.async
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+class ScalaFutureTracingService extends AsyncTracingService {
+  implicit private val context: ExecutionContext = ExecutionContext.global
+
+  override def supports(future: AnyRef): Boolean = future != null && future.isInstanceOf[Future[_]]
+
+  override def whenDone(value: AnyRef, f: (Any, Throwable) => Unit): Future[_] = {
+    value.asInstanceOf[Future[_]].transform(result => {
+      f(result, null)
+      result
+    }, throwable => {
+      f(null, throwable)
+      throwable
+    })
+  }
+}


money-aspectj/src/test/scala/com/comcast/money/aspectj/TraceAspectSpec.scala
```diff
@@ -76,6 +79,12 @@ class TraceAspectSpec extends WordSpec
     throw new RuntimeException("not-ignored")
   }
 
+  @Traced(
```
pauljamescleary: As opposed to adding an async flag, with aspectj anyway we can add a different joinpoint / pointcut on functions with annotations `@Trace` that return a promise (I think).  Have you tried that?  Unsure off the top of my head if that would work

HaloFour: That would probably work although might require some effort to avoid ambiguity between the existing pointcut.  The problem with that is that it locks it to supporting only specific types and there is a fairly wide ecosystem of these types out there since Java didn't add one offering a callback until Java 8.  `Promise[T]` is quite specific to Scala and its uncommon to see in other Java libraries.

pauljamescleary: Right, I am unclear if it would work.  Can we have a test where, if someone sets async to be true, on something that is not async, that everything still works?

Imagine

```
@Traced(value="sync", async=true)
def plusOne(x: Int): Int = x + 1
```

HaloFour: Agreed, the tests are a little anemic right now.  This PR is more a spike for the purpose of discussion.

I might need to approach the unit tests a little differently as they should validate that log records aren't written until the promise/future is complete and the current helper methods don't really lend themselves to that, short of waiting for a timeout to expire.  I'm thinking of mocking the Tracer instead.

money-core/src/main/scala/com/comcast/money/core/async/AsyncTracingService.scala
```diff
@@ -24,7 +24,7 @@ object AsyncTracingService {
   private lazy val services = loadServices()
 
   def findTracingService(future: AnyRef): Option[AsyncTracingService] =
-    services.find(_.supports(future))
+    if (future != null) services.find(_.supports(future)) else None
```
pauljamescleary: nitpicky, but...

```
Option(future).map(f => services.find(_.supports(f))
```

HaloFour:
I've reworked the asynchronous tracing aspect so that it behaves a bit differently.

Two spans are created instead of one.  The first span represents the synchronous execution of the method.  The second span, which is a child of the first span, represents the asynchronous operation represents by the future returned from the method.  The second span is popped from the stack of spans and is left open after the method returns.  When the future is complete, that triggers the second span to be closed based on the result of the future.

That second span is never closed if the following situations happen:

1.  The method returns `null`.
2. The method returns something that isn't recognized to be a future
3. The method throws an exception

The success of the second span depends on the success of the future and respects the same ignored exceptions.

Given that the second asynchronous span is popped from the `SpanLocal` would there be any negative consequences from never stopping it?

money-core/src/main/scala/com/comcast/money/core/async/AsyncTracingService.scala
HaloFour: nitpicky back at you, but:  😁 

```scala
Option(future).flatMap(f => services.find(_.supports(f))
```

HaloFour:
I'll also note that I don't particularly care for the name `AsyncTracingService`, but I haven't thought of a good alternative.  I didn't want to waste too much time on it and I'm open to suggestions.

money-aspectj/src/main/scala/com/comcast/money/aspectj/TraceAspect.scala
```diff
@@ -39,22 +42,70 @@ class TraceAspect extends Reflections with TraceLogging {
...
         throw t
     } finally {
+      tracer.stopSpan(result)
       mdcSupport.setSpanNameMDC(oldSpanName)
+    }
+  }
+
+  private def traceMethodAsync(joinPoint: ProceedingJoinPoint, traceAnnotation: Traced): AnyRef = {
+    val key = traceAnnotation.value()
+    val asyncKey = key + "-async"
```
pauljamescleary: why do we append -async to the name?

HaloFour: I'm creating two spans.  I'm giving the second span which represents the async task the suffix `-async`.  I could change that to do something else, or make it configurable.

pauljamescleary: Ok, that is where I was confused.  I don't understand why we need two separate spans.

We use Money heavily with scala futures, and all tracing is declaritive in the code.  We want to give the user the ability to start / stop however we want.

It is a little unclear here with the `@Traced` annotation as to when the span actually starts.  The way we have done this is that the span starts when the method is invoked and completes when the Future (Promise) ends.  Something conceptually like this...

```
def doSomethingTraced(foo: String): Future[String] = {
  val span = startSpan("myName")
  val fut = doSomethingInTheFuture(foo)
  fut.onComplete { case Success(x) => span.stop(true); case Failure(e) => span.stop(false) }
}
```

HaloFour: I see.  That is the approach I was taking in the first iteration and I have no problem going back to that model.  If I do that then the code above could be reduced to:

```scala
@Traced(value = "myName", async = true)
def doSomethingTraced(foo: String): Future[String] =
    doSomethingInTheFuture(foo)
```

HaloFour:
Okie dokie, I've reverted the logic back to the previous implementation, where a single span is created prior to the annotated method being called and that span not being stopped if the method returns a recognized future that is not yet complete.  I've updated all of the unit tests to also reflect this change.

pauljamescleary:
Thumbing this up.  Thanks for the contribution!!!

Let me know if there are any other pending modifications to this, and if I can merge this before your other PR.

HaloFour:
Great.  I am sitting on a couple of additional unit tests but no modification to the current functionality.

I was toying with the idea of switching from using the Java ServiceLoader to instead using the same typesafe Config mechanism that Money uses for handlers and other configuration.  I reached for ServiceLoader since I was more familiar with it coming from Jersey and it's frequently used for pluggable connectors like this but I'd want it to fit in with the established paradigm used by Money.  Kris didn't have an opinion either way.

pauljamescleary:
Good to merge @HaloFour ?

money-aspectj/src/test/scala/com/comcast/money/aspectj/TraceAspectSpec.scala
```diff
@@ -146,6 +204,141 @@ class TraceAspectSpec extends WordSpec
...
+        Then("the async execution is logged")
+        expectSpanInfoThat("is named asyncMethodReturnsPromise and is not successful", span =>
+          span.name == "asyncMethodReturnsPromise" && !span.success())
+      }
+      "handle async methods that return a failed future with ignored exception" in {
+        Given("a method that returns a future")
+
+        When("the method is invoked")
+        val promise = Promise[String]()
+        val future = asyncMethodWithIgnoredException(promise)
```
ptravers: should `future` maybe be `promise` I feel like I'm being lied to

HaloFour:
Refactored to using typesafe Config to load the `AsyncNotificationHandler` instances (and changed their name to fit better with `SpanHandler`).  Refactoring some bits to be more idiomatic Scala (with some help from @ptravers who knows Scala much better than I).

Open questions:

1.  The `AsyncNotificationHandler` returns a value of `AnyRef`.  This is to allow the handler to replace the return value with another compatible return value, just in case we'd want our tracing callback to happen before other registered callbacks.  This does complicate the implementation and may not be that important.

2.  I have implemented my own `ExecutionContext` which invokes the callback on the current thread.  While this is generally frowned upon the thought is that the callback to finish tracing should be very fast and it would be a waste of resources and could potentially bloat the span times to put that back onto the global ExecutionContext.  I expose the `ExecutionContext` parameter through the `AsyncNotificationHandler` trait, which is probably a bit overkill, and I could probably isolate that to just the `ScalaFutureNotificationHandler` class instead.

money-core/src/main/scala/com/comcast/money/core/async/AsyncNotificationHandlerFactory.scala
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+
+package com.comcast.money.core.async
+
+import com.typesafe.config.Config
+
+object AsyncNotificationHandlerFactory {
+
+  def create(config: Config): AsyncNotificationHandler = {
```
ptravers: What about `object AsyncNotificationHandler` and `create` as `apply` would that look bad at call site? I guess this way you can just `import AsyncNotificationHandlerFactory.create`

money-core/src/test/scala/com/comcast/money/core/SpecHelpers.scala
```diff
@@ -27,14 +27,25 @@ import scala.collection.{ Set, mutable }
...
   private val messages = new mutable.HashMap[String, mutable.Set[String]] with mutable.MultiMap[String, String]
 
-  def clear() = messages.clear()
+  def clear(): Unit = {
+    messages.clear()
+    spans.clear()
+  }
 
-  def add(log: String, message: String) = messages.addBinding(log, message)
+  def add(log: String, message: String): Unit = {
```
ptravers: I kind of like it better without the brackets but...

HaloFour:
I ended up removing the `ExecutionContext` parameter.  All it really did was add a lot of noise and the interface would only really make sense in Scala-specific contexts.  I still use the `DirectExecutionContext` directly within `ScalaFutureNotificationHandler` to keep the span callback on the same thread.

HaloFour:
Okay, I lied.  The complexity of `TraceAspect#adviseMethodsWithTracing` bothered me so I refactored the async portion out into a helper method with lots of comments and I refactored a bit of the method itself to hopefully be a little more clear.

Now I think I'm good to merge this PR.  😀

money-aspectj/src/test/scala/com/comcast/money/aspectj/TraceAspectSpec.scala
pauljamescleary: The main test that I would like to try out is nested futures (futures with flatmaps).

When I worked with scala Futures in the past, the onComplete of a future was executed in odd ways.

Imagine you have 3 levels of functions / nested futures via flatmaps and maps.  You create child spans at each level, so 3 levels deep in a span tree.  The onComplete and context that was passed was "off".

Sorry I don't have more information, I have to check out your PR and try and re-create the issue I saw

HaloFour: You mean something like this?

```scala
@Traced(value = "asyncGetHello", async = true)
def asyncGetHello(): Future[String] = Future { "hello" }
@Traced(value = "asyncGetWorld", async = true)
def asyncGetWorld(): Future[String] = Future { "world" }

@Traced(value = "someAsyncMethod", async = true)
def someAsyncMethod(): Future[String] = {
    for {
        hello <- asyncGetHello()
        world <- asyncGetWorld()
    } yield s"$hello $world!"
}
```

HaloFour: I think I see what you mean.  While the above does produce three spans as expected, the parent of span "asyncGetWorld" is not "someAsyncMethod".  Is that the behavior that you experienced?

HaloFour: So, assuming that this is the problem to which you were referring, I've thought of a way in which it can be resolved, although it's a little heavy-handed.  The issue is that we do not control the thread on which `asyncGetWorld` is invoked.  At best we can hand a custom `ExecutionContext`, such as the `TraceFriendlyExecutionContextExecutor` already in `money-core`.  The problem is also that the thread on which `ExecutionContext.execute` is invoked is the thread on which the `Future` happened to complete, which is not the thread that we control and it's too late to capture a span.  The best we can do here is to have `TraceFriendlyExecutionContextExecutor` capture a span as soon as it is created and to apply that span on every `Runnable` submitted to `execute`.  But this requires that a new `TraceFriendlyExecutionContextExecutor` be created at the beginning of every single async method and prevents importing a common executor.

```scala
@Traced(value = "asyncGetHello", async = true)
def asyncGetHello(): Future[String] = Future { "hello" }
@Traced(value = "asyncGetWorld", async = true)
def asyncGetWorld(): Future[String] = Future { "world" }

@Traced(value = "someAsyncMethod", async = true)
def someAsyncMethod(): Future[String] = {
    implicit val ec: ExecutionContext = TraceFriendlyExecutionContext(ExecutionContext.global, SpanLocal.current)
    for {
        hello <- asyncGetHello()
        world <- asyncGetWorld()
    } yield s"$hello $world!"
}
``` 

We could probably make that less verbose, but it still puts a lot of onus on the developer to get right.

Alternately, I have tested an aspect which intercepts all calls to `Future` methods which accept an `ExecutionContext` within a method annotated with `@Traced` which wraps the `ExecutionContext` with one that immediately captures the span.  This seems to work pretty well, but as I mentioned, is quite heavy handed.

A third option might be to have an aspect that intercepts all calls to any method that can return a `Future` within a method annotated with `@Traced` and wrap the `Future` itself.  I don't think that's much different and may prove more problematic for code that might extend from `Future` directly.

pauljamescleary: In the version 0.8.x codebase, I had written a `TracedFuture` that handled this well.  I abandoned it, maybe I shouldn't have.  I will dig that up.  There is also a unit test that tests out the proper behavior

pauljamescleary: Here is the traced future stuff - https://github.com/Comcast/money/blob/0.8.x/money-core/src/main/scala/com/comcast/money/concurrent/Futures.scala

pauljamescleary: Here are the unit tests that illustrate the problem ....

https://github.com/Comcast/money/blob/0.8.x/money-core/src/test/scala/com/comcast/money/concurrent/FuturesSpec.scala

pauljamescleary: I would be interested to test out the second approach you mentioned.  The code I had extended Futures, but iirc, not all use cases were cleanly supported.  Plus, we have this TracedFuture thing.

HaloFour: I've pushed a second branch to my fork of the repo that contains this aspect and the simple test above.

https://github.com/HaloFour/money/blob/async-tracing-future-aspect/money-aspectj/src/main/scala/com/comcast/money/aspectj/FutureAspect.scala

HaloFour: I've been playing around a little with the `TracedFuture[T]` code and unit tests that you linked, and I think it runs into a fundamental problem when it comes to the lifespan of the span.  In your example you are stopping the span as soon as the `TracedFuture[T]` completes, but that happens before any composed operations are executed, so you'd lose recording any data to the span from those callbacks.  But if you leave the span open there is no way to know when the span should be stopped.  You'd need a kind of terminal operator which the developer would be required to call at the end of their composition.

Rather than going to that trouble it might be easier to expose a helper method in which you'd wrap all of the future logic:

```scala
object Futures {
  private val tracer = Money.Environment.tracer

  def traceFuture[T](key: String)(f: (ExecutionContext) => Future[T])(implicit executionContext: ExecutionContext): Future[T] = {
    val span = SpanLocal.current match {
      case Some(existingSpan) => tracer.spanFactory.childSpan(key, existingSpan)
      case _ => tracer.spanFactory.newSpan(key)
    }

    val future = f(wrap(executionContext, span, captureMdc()))

    future onComplete {
      case Success(_) => span.stop(true)
      case _ => span.stop(false)
    }

    future
  }
}
```

then:

```scala
traceFuture("root") { implicit ec =>
  tracer.record("begin", "root")
  println(s"\r\n; begin root ${SpanLocal.current}")

  traceFuture("nested") { _ =>
    tracer.record("begin", "nested")
    println(s"\r\n; begin nested ${SpanLocal.current}")
    Future { Some(456) }
  }.flatMap {
    case Some(v) =>
      tracer.record("flatMap", "nested")
      println(s"\r\n; flatMap nested ${SpanLocal.current}")
      Future {
        v
      }
  }.map {
    _ =>
      tracer.record("map", "nested")
      println(s"\r\n; map nested ${SpanLocal.current}")
      SpanLocal.current.get
  }
}.map { child =>
  tracer.record("flatMap", "root")
  println(s"\r\n; flatMap root ${SpanLocal.current}")
  child
}.map {
  s =>
    tracer.record("map", "root")
    println(s"\r\n; map root ${SpanLocal.current}")
    (SpanLocal.current.get, s)
}
```

pauljamescleary: Right, I recall now that we did not bring the TracedFuture forward due to stopping spans.

The old 0.8.x tracing was based on actors, and spans would "hang out" for a bit before being released.  This allowed the span to be stopped multiple times, so the TracedFuture stuff worked ok.

We do not do that in the current 0.9.x branch.

I was thinking of using a monad transformer based around StateT.  The cedi guys did something similar in cedi-dtrace; however there design was based on ReaderT, which did not allow flexible modification of the Span in flight.

https://github.com/ccadllc/cedi-dtrace

Would love to build on the cedi stuff, use StateT instead.

I wonder if we had that, would the trace aspect stuff become easier?  Like, we could wrap a Future in a StateT at certain points, and then pull it out of StateT?

HaloFour: I'll have to spend some time grokking over cedi and cats to get a better understanding of it.  I'm a bit of a Scala neophyte at the moment.  My concern is what this would look like over in Java-land where the Scala sugar would be absent.  How much onus does this put on the developer to explicitly spell out their tracing?  How would it work over other types?

Most of the complexity of the additions to `TraceAspect` here are in my attempt to be future-agnostic, hoping that it would eventually support some of the several abstractions ecosystem, particularly Java 8 `CompletionStage`, Java 9 `Flow`, RxJava and Spring Reactor, all of which work a little differently.  I reached for Scala `Future` here because it was already there.  It works well enough for that purpose, as long as you don't compose of the future within the annotated method.  The implementation details would be different for each abstraction, though.

I was able to greatly simplify the [`FutureAspect`](https://github.com/HaloFour/money/blob/async-tracing-future-aspect/money-aspectj/src/main/scala/com/comcast/money/aspectj/FutureAspect.scala) I was playing with in the other branch so that it only required two pointcuts and a single advice.  It works well with or without the `@Trace` aspect.  I need to write up some tests to prove it out.

pauljamescleary: I think if you can manage to get those old trace unit tests working properly, we should be good.

I really like the idea of extending this to other async metaphors too, that is great.

Sorry about the slow turnaround on this stuff.  I have been swamped lately.  Busy clearing things off my plate so I can dive into this deeper.

HaloFour: No worries.  I've been pretty swamped too.

I've taken your tests and made the majority of them work using TraceAspect and FutureAspect.  The major ones that are still failing are those that use the `onSuccess` and `onFailure` methods as they don't return new `Future` instances which creates a race condition between the callback logging to the span and the span being closed.  I don't have a good answer as to how to solve that yet.

HaloFour:
I've commited the `FutureAspect` and converted tests.  Most of them pass but a few are kind of flaky and I'm still looking into the reason why.

There are three tests which fail regularly and I don't have a good answer for them.  They rely on terminal/non-composed methods on `Future[T]` which create an inherent race condition between their execution and the span being stopped due to the future being completed.  To my knowledge there is no prescribed order in which the registered callbacks of a future are executed and those callbacks are subsequently submitted to an execution context which makes the order of operations even more nondeterministic.  I'll keep poking at ideas but I have a feeling that such operations wouldn't really compose well into a span.

The methods in question are:

```
onComplete
onSuccess
onFailure
foreach
```

money-aspectj/src/test/scala/com/comcast/money/aspectj/TraceAspectSpec.scala
pauljamescleary: I wonder if we can just have a different type of `Span` created different than `CoreSpan`...one that can accept multiple stops.

In a really high throughput system if you sample EVERY request, that can be tough.  One of the biggest misses so far with money is the lack of sampling.  Basic sampling strategies are so simple to implement I kick myself for not doing it years ago.  If we had some level of sane sampling, we wouldn't have to worry too much about memory pressure.

Imo, we can partially close a span on the first stop, and then kinda wait a little before releasing the span info for handling.  With a sane sampling rate that is high (like 100 samples per second), those spans could easily hang out for a minute or so.

pauljamescleary: Also, I think your example is odd.  Seems like you have a future that yields a future as a value.  That is dumb (not calling you dumb, but that doesn't seem like a good use case).  I am curious if some of the contrived tests I wrote are things that people would never write?

Certainly `.flatMap { x => Future { x + 1 } }` is valid

But `Future { Future { x }  }` is not

HaloFour: Which example is that?  I'm trying to duplicate (or at least reinterpret) the unit tests you had around traced futures.  Some of the examples are contrived intentionally but I also wouldn't discount my flubbing things up.  I've not had the luxury of dedicating my attention to this recently.

HaloFour: I was playing with going down a path of having `TraceAspect` replace the existing `Span` with a wrapper `AsyncSpan` which internally would use an `AtomicInteger` as a reference counter.  Invoking a terminal method on `Future` like `onComplete` would bump that counter up.  When that callback would complete it would call `stop`, which would decrease the counter, as would when the instrumented future completed.  Only when the counter reached zero would the span actually be stopped.  I don't think that's quite what you were talking about, though.

pauljamescleary: That’s a great idea!

pauljamescleary: I was thinking that perhaps my old TraceAspect unit tests had test cases that were invalid.

HaloFour:
I wonder if you would be amenable to breaking this into separate PRs.  I see two orthogonal concerns here:

1.  Support the tracing of methods that return a value representing a "future" where the span remains open until that "future" is complete.
1.  Support "capturing" and "continuing" the current `Span` on the callbacks registered to a particular "future" so that whatever they trace can be correlated together.

The former I think is pretty cut and dry.

The latter gets into the weeds of implementation details around each implementation of "future", as we're seeing with `scala.concurrent.Future`.  We'd probably run into the same kinds of problems trying to extend that support to `java.util.concurrent.CompletionStage`, `com.google.common.util.concurrent.ListenableFuture`, `rx.Observable` or any of the other potential libraries that exist out there.

pauljamescleary:
I think that cutting this up makes sense.

HaloFour:
I've reset this branch back to the commit prior to diving into the `Future` rabbit hole.  If a method returns a `Future` and is annotated with `@Traced(value = "whatever", async = true)` then the `Span` will not be closed until after that `Future` is completed.  Additional handlers can be configured to support other future types also, like `java.util.concurrency.CompletionStage`.

Speaking of, I'd love to make that a module of this project but it looks like Money currently targets Java 6.  Is that to retain compatibility with older projects or would it be possible to bump the version up to 8?

PR #69, opened 2016-05-04
Actor support

derjust:
This is the first part of a PR to support Akka's actor system with money

The first step is to support a consistent storage for span's in the context of actors exchanging messages.
This means spans has to be started and stopped manually as depicted in the test case `MoneyExtensionSpec`.

Further PRs will add convenience methods to support automatic starting & stopping for messages flowing through the actor system

PR #89, opened 2018-03-20
Akka stream tracing

ptravers:
Commit adds a package that can be used to pass Spans through an Akka Stream.

As part of this the concept of a SpanContext had to be introduced this was taken from #69 

Some of this code was copied across from that PR.


project/MoneyBuild.scala
```diff
@@ -285,6 +303,14 @@ object MoneyBuild extends Build {
     val codahaleVersion = "3.0.2"
     val apacheHttpClientVersion = "4.3.5"
 
+    val akkaV = "2.4.14"
```
pauljamescleary: are we sure about these akka and akka-http versions?  Not sure if they are the current

money-akka/src/main/scala/com/comcast/money/akka/TracedFlow.scala
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  def traceStageAndStop(key: String, stageLogic: In => Out, isSuccessful: Boolean = true): Unit = {
+    implicit val (inMessage, spanContext): (In, SpanContextWithStack) = grab[TracedIn](in)
+    tracer.startSpan(key)
+    val outMessage = stageLogic(inMessage)
+    push[TracedOut](out, (outMessage, spanContext))
+    spanContext.getAll.foreach(_ => tracer.stopSpan(isSuccessful))
+  }
+}
```
pauljamescleary: needs a new line

pauljamescleary:
before we merge this, we also need to build the license headers.

Also, please sign the CLA :)

project/MoneyBuild.scala
ptravers: Current would be `akkaV = 2.5.11` and `akka-http = 10.1.0` but we aren't running that as of right now on PAPI. Not sure what the effect would be of having money depend on the latest version. I can take a look.

money-akka/src/main/scala/com/comcast/money/akka/TracedFlow.scala
ptravers: Sure

ptravers:
Hey yeah I would be happy to be the maintainer for it. That's not a problem.

PR #92, opened 2018-04-11
Tracing combinators first steps

ptravers:
The goal of this branch is to build a way to incorporate generic Flows in to a traced stream. After the previous PR for this package it was only possible to do so by constructing a Flow using Graph Stage Logic.

example "generic" Flows:
`Flow[Int]`
`Flow[Int, String].fromFunction(_ => "a")` 

A common stream use case which is currently impossible to instrument without converting every segment in the stream to instead take `(Int, SpanContextWithStack)` and adding additional Flows to start and stop Traces
```
Source(1 to 3) ~> Flow[Int] ~> Sink.seq[Int]
```

This PR enables the below syntax instead producing a traced stream:

```
(Source(1 to 3) |~> Flow[Int]) ~| Sink.seq[Int].in
```
This will start and stop a Span for the entire stream and also start and stop a span for the Flow. These two would to all intents and purposes at the moment contain the same stats most likely as the Source and Sink are not traced.

The current implementation will always provide a trace named Stream which will pass through all the Flows in a stream to the Sink.

Current Operators and their meanings (very much looking for input on this):
`|~>` trace this next element and continue to pass the SpanContext
`~|` close trace and return Unit

`~|>` close trace after the next element and continue the stream
example:

```
val out = (source |~> Flow[String]) ~|> Flow[String]

SourceShape(out.outlet)
```

`~\` choose the first inlet of a fan in
`~/` choose the second inlet of a fan in

example:

```
Source(List("chunk", "funk")) |~> partition

partition.out(0) |~> Flow[String] |~\ concat

partition.out(1) |~> Flow[String] |~/ concat

concat ~| sink.in
```

Current issues there are probably too many different types of combinators. The only necessary ones are probably `|~>`, `~|` and `~|>`. Although it's pretty debatable about the `~|`. I do want the user of the combinators to understand when a combinator is stopping the trace of the stream and they can either return to using `~>` or they can return their Shape.


ptravers:
Currently it feels very cumbersome and that possibly there are too many tools in use. For instance the TraceBuilder is necessary to be able to take FanIn and FanOut Shapes and make them traced but possibly the whole thing should really run through the TracedBuilder

money-akka/src/main/scala/com/comcast/money/akka/stream/TracedFlow.scala
```diff
@@ -39,9 +42,11 @@ abstract class TracedLogic[In, Out](shape: Shape)
     tracer.startSpan(key)
     val outMessage = stageLogic(inMessage)
     push[TracedOut](out, (outMessage, spanContext))
+    tracer.stopSpan()
```
pauljamescleary: why is this stopping twice?

money-akka/src/main/scala/com/comcast/money/akka/stream/TracedStreamCombinators.scala
```diff
@@ -0,0 +1,171 @@
...
+
+  import DefaultSpanKeyCreators._
+  import akka.stream.scaladsl.GraphDSL.Implicits._
+
+  implicit class PortOpsSpanInserter[In: ClassTag](portOps: PortOps[(In, SpanContextWithStack)])
+                                                  (implicit builder: Builder[_],
+                                                   fskc: FlowSpanKeyCreator[In] = DefaultFlowSpanKeyCreator[In]) {
+    type TracedIn = (In, SpanContextWithStack)
+
+    def ~|>[Out: ClassTag](flow: Flow[In, Out, _]): PortOps[(Out, SpanContextWithStack)] =
```
pauljamescleary: unclear if these make sense when called, but in general when using symbols for operators, I recommend that they are aliases for another function with a more descriptive name.

Also, really need scaladoc on all of these things

Like if `~|>` means `injectFlow` there should be a function called `injectFlow`

money-akka/src/main/scala/com/comcast/money/akka/stream/TracedFlow.scala
ptravers: Yeah so I made a mistake first time through which was to only close all the spans at the very end. Which I had thought from the previous PR was what was wanted but after discussing with Kris this made no sense as it was not actually tracking the time taken in this section of the flow. I will have to change the `traceStageAndStopAll` to make more sense.

money-akka/src/main/scala/com/comcast/money/akka/stream/TracedStreamCombinators.scala
ptravers: I guess on the making sense part it's primarily directed at people who already use GraphDSL but yeah I totally get what you mean.

Sure I absolutely can do on converting in to an underlying named function.

hundred percent on the scala doc will do now.

money-akka/src/main/scala/com/comcast/money/akka/stream/TracedFlow.scala
ptravers: I just noticed the duplication yeah I fixed that just now

pauljamescleary:
This is impressive

PR #93, opened 2018-04-18
Log json for elk

MKnowl001:


money-core/src/main/scala/com/comcast/money/core/handlers/StructuredLogSpanHandler.scala
```diff
@@ -0,0 +1,81 @@
...
+      ("host", spanInfo.host()),
+      ("start-time", spanInfo.startTimeMillis()),
+      ("end-time", spanInfo.endTimeMillis()),
+      ("span-duration", spanInfo.durationMicros()),
+      ("span-success", spanInfo.success())
+    )
+    val noteFields: Seq[(String, Any)] = spanInfo.notes.values.asScala.map(n => (n.name(), n.value())).toSeq
+    val allFields = baseFields ++ noteFields
+
+    allFields.foreach(p => MDC.put(p._1, p._2.toString))
```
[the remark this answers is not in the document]
MKnowl001: When you write to MDC it makes all the fields available to the logging system as fields. If they are just part of the message, the logging system has no knowledge of them. Once they are in MDC, you can configure logging to use them. For example, in logback.xml you can define your logged output with `<pattern>... trace id: %X{trace-id}, span id:  %X{span-id} etc </pattern>`.   For logging to ELK, this encoder ` <encoder class="net.logstash.logback.encoder.LogstashEncoder"/>` finds the fields in MDC and they are propagated into elastic search.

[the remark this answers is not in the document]
MKnowl001: I found a mention of MDC here https://github.com/Comcast/money/wiki/Money-Configuration#emitters in "Default Configuration" , but the only place I see it used is https://github.com/Comcast/money/blob/master/money-akka/src/main/scala/com/comcast/money/akka/MoneyExtension.scala ... to say that the akka component will ignore it!

money-core/src/test/scala/com/comcast/money/core/handlers/StructuredLogSpanHandlerSpec.scala
```diff
@@ -0,0 +1,165 @@
...
+import org.scalatest.mock.MockitoSugar
+import org.scalatest.{Matchers, OneInstancePerTest, WordSpec}
+import org.slf4j.Logger
+
+class StructuredLogSpanHandlerSpec extends WordSpec
+    with Matchers with MockitoSugar with OneInstancePerTest with TestData {
+
+  val mockLogger = mock[Logger]
+  val sampleMessage = "sample formatted log message"
+//  val sampleFormatterConfig = ConfigFactory.parseString("formatting { this=that }")
```
pauljamescleary: remove comment

money-core/src/test/scala/com/comcast/money/core/handlers/StructuredLogSpanHandlerSpec.scala
```diff
@@ -0,0 +1,162 @@
...
+          |formatting {
+          | this=that
+          |}
+        """.stripMargin
+      )
+
+      underTest.configure(config)
+      underTest.handle(testSpanInfo)
+
+      verify(mockLogger).error(logEntryCaptor.capture())
```
pauljamescleary: since you are not using the result of the captor, you can use the `org.mockito.Matchers._` matchers...

`verify(mockLogger).error(anySring)` ... they have a bunch of `any` matchers to use.

money-core/src/test/scala/com/comcast/money/core/handlers/StructuredLogSpanHandlerSpec.scala
```diff
@@ -0,0 +1,162 @@
...
+          |formatting {
+          | this=that
+          |}
+        """.stripMargin
+      )
+
+      underTest.configure(config)
+      underTest.handle(testSpanInfo)
+
+      verify(mockLogger).warn(logEntryCaptor.capture())
```
pauljamescleary: I would make the same changes here and elsewhere as above

PR #95, opened 2018-04-27
Add support for X-B3 headers

MKnowl001:


money-core/src/main/scala/com/comcast/money/core/Formatters.scala
```diff
@@ -35,4 +35,12 @@ object Formatters {
 
   def toHttpHeader(spanId: SpanId): String =
     HttpHeaderFormat.format(spanId.traceId, spanId.parentId, spanId.selfId)
+
+  def fromB3HttpHeaders(traceId: String, maybeParentSpanId: Option[String], maybeSpanId: Option[String]) = Try {
+    (maybeParentSpanId, maybeSpanId) match {
```
pauljamescleary: unclear as to why this is in a `Try`

pauljamescleary: make sure to define an explicit return type.

Also, I think `Try` isn't correct, we want this to return an `Option`

money-java-servlet/src/test/scala/com/comcast/money/java/servlet/TraceFilterSpec.scala
```diff
@@ -52,21 +52,124 @@ class TraceFilterSpec extends WordSpec with Matchers with OneInstancePerTest wit
...
+      when(mockRequest.getHeader("X-B3-SpanId"))
+        .thenReturn(existingSpanId.selfId.toString)
+
+      underTest.doFilter(mockRequest, mockResponse, mockFilterChain)
+
+      SpanLocal.current.value.info.id shouldEqual existingSpanId
+    }
+
+
+
```
ptravers: unnecessary white space one blank line should suffice between tests

money-java-servlet/src/test/scala/com/comcast/money/java/servlet/TraceFilterSpec.scala
```diff
@@ -52,21 +52,124 @@ class TraceFilterSpec extends WordSpec with Matchers with OneInstancePerTest wit
...
+        .thenReturn(existingSpanId.selfId().toString)
+
+      underTest.doFilter(mockRequest, mockResponse, mockFilterChain)
+
+      SpanLocal.current.value.info.id.traceId() shouldEqual existingSpanId.traceId()
+      SpanLocal.current.value.info.id.selfId() should not equal(existingSpanId.selfId())
+    }
+
+
+
```
ptravers: excess whitespace

money-java-servlet/src/test/scala/com/comcast/money/java/servlet/TraceFilterSpec.scala
```diff
@@ -76,6 +179,48 @@ class TraceFilterSpec extends WordSpec with Matchers with OneInstancePerTest wit
...
+        "X-B3-TraceId", existingSpanId.traceId
+      )
+      verify(mockResponse).addHeader(
+        "X-B3-ParentSpanId", existingSpanId.parentId().toString
+      )
+      verify(mockResponse).addHeader(
+        "X-B3-SpanId", existingSpanId.selfId().toString
+      )
+    }
+
```
ptravers: excess whitespace

money-core/src/test/scala/com/comcast/money/core/FormattersSpec.scala
```diff
@@ -22,11 +22,35 @@ import org.scalatest.{ Matchers, WordSpec }
...
+
+      val actualSpanId = Formatters.fromB3HttpHeaders(expectedSpanId.traceId, Option(expectedSpanId.parentId().toString),Option(expectedSpanId.selfId().toString)).get
+      actualSpanId shouldBe expectedSpanId
+    }
+
+
+    "convert from 2 X-B3  http headers" in {
+
+      val actualSpanId = Formatters.fromB3HttpHeaders("a", Option("1"), None).get
+      actualSpanId.traceId shouldBe "a"
```
ptravers: I would prefer if this was a closer representation of what a B3 Header looks like.

money-java-servlet/src/main/scala/com/comcast/money/java/servlet/TraceFilter.scala
```diff
@@ -33,35 +33,69 @@ import scala.util.{ Failure, Success }
...
 
+  private val spanName = "servlet"
+
   override def doFilter(request: ServletRequest, response: ServletResponse, chain: FilterChain): Unit = {
 
     SpanLocal.clear()
     val httpRequest = new HttpServletRequestWrapper(request.asInstanceOf[HttpServletRequest])
-    val incomingTraceId = Option(httpRequest.getHeader(MoneyTraceHeader)) map { incTrcaceId =>
+
+    val maybeMoney = Option(httpRequest.getHeader(MoneyTraceHeader))
```
pauljamescleary: I think that the logic is a little hard to follow here.

The process should be:

1. Attempt to get parse the money header, fall back to the b3 header
2. Set the SpanLocal if either of the headers is present
3. Always just force copy over the request header to the response header.

```scala
def parseMoneyTraceHeader(httpRequest: HttpRequest): Option[SpanId] = ???

def parseB3TraceHeader(httpRequest: HttpRequest): Option[SpanId] = ???

```

With those two helpers, we can attempt to pull out the passed in span id

```scala
val spanId: Option[SpanId] = parseMoneyTraceHeader(httpRequest).orElse(parseB3TraceHeader)
spanId.foreach(s => SpanLocal.push(factory.newSpan(s, spanName)))
```

Then, on the way out, all we need to do is duplicate the request header...

```scala
Option(httpRequest.getHeader(Money)).foreach(response.setHeader(...))
Option(httpRequest.getHeader(B3TraceId)).foreach(response.setHeader))
Option(httpRequest.getHeader(B3ParentId)).foreach...
Option(httpRequest.getHeader(B3SpanId)).foreach...

// given the repetition above, would probably make sense to have a simple helper function like 

def setResponseHeader(headerName): Unit = Option(httpRequest.getHeader).foreach(httpResponse.setHeader...)

```

money-core/src/main/scala/com/comcast/money/core/Formatters.scala
```diff
@@ -35,4 +35,12 @@ object Formatters {
...
+      case (Some(ps), Some(s)) => new SpanId(traceId, ps.toLong, s.toLong)
+      case (Some(ps), _) => new SpanId(traceId, ps.toLong)
+      case _ => new SpanId(traceId)
+    }
```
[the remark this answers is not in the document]
MKnowl001: toLong can throw a NumberFormatException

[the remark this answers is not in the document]
pauljamescleary: Eww, good catch.

A little complicated the situation:

1. You may have the TraceId header, but not the parent id or the span id
2. You may have the TraceId header, the parent id header, but not the span id
3. You may have all 3
4. You may have none
5. The parent id header could be invalid
6. The span id header could be invalid

I do think Try is appropriate (we have not upgraded to Either yet).  We do need to encode all those rules though.

The signature should take 3 options (traceId: Option[String], parentId: Option[String], spanId: Option[String]) and return a Try[SpanId].

We should fail if any of the `toLong` fails.  Otherwise, we should always return a Success with a SpanId in it.

I tried to play with an implementation but could not get there today.  It is a little tough to deal with, and perhaps just wrapping the whole thing in a `Try` is good enough.

money-java-servlet/src/main/scala/com/comcast/money/java/servlet/TraceFilter.scala
pauljamescleary: Looks like we may keep the `Try`, in which case we can still follow this.  The `parseMoneyTraceHeader` would attempt to parse (if not there it would be none).  If there, get the try, log any errors, and just do a `toOption`.

We do want to log I think, but not fail the request.

money-core/src/main/scala/com/comcast/money/core/Formatters.scala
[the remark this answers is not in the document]
pauljamescleary: I do not have a much better solution, I think this is good, changing my comment

pauljamescleary:
nice work!  A few small things and I think I am good.

Let's see what @ptravers thinks

money-core/hack.sc
```diff
@@ -0,0 +1,48 @@
...
+def test(s1: Option[String],s2: Option[String]) = (s1,s2) match {
+  case (Some(x1),Some(x2)) if(s2.nonEmpty) => print("a ")
+  case (Some(s),_) => print("b ")
+  case _ => print("c ")
+}
+
+test(a,c)
+b.getOrElse("xxx")
+
+
```
ptravers: Can this be removed please?

money-java-servlet/src/main/scala/com/comcast/money/java/servlet/TraceFilter.scala
```diff
@@ -16,52 +16,94 @@
...
+
+  def parseMoneyTraceHeader(httpRequest: HttpServletRequest): Option[SpanId] = {
+
+    def parseHeader(headerValue: String): Option[SpanId] = {
+      fromHttpHeader(headerValue) match {
+        case Success(spanId) => Some(spanId)
+        case Failure(ex) => {
+          logger.warn("Unable to parse money trace for request header '{}'", headerValue)
+          None
+        }
```
ptravers: would personally prefer if the braces were dropped after the case

money-java-servlet/src/main/scala/com/comcast/money/java/servlet/TraceFilter.scala
```diff
@@ -16,52 +16,94 @@
...
+      fromB3HttpHeaders(traceIdVal, maybeParentSpanId, maybeSpanId) match {
+        case Success(spanId) => Some(spanId)
+        case Failure(ex) => {
+          logger.warn(s"Unable to parse X-B3 trace for request headers: " +
+            s"${B3TraceId}:'${traceIdVal}', " +
+            s"${B3ParentSpanId}:'${maybeParentSpanId.getOrElse("None received")}', " +
+            s"${B3SpanId}:'${maybeSpanId.getOrElse("None received")}'") +
+            s"${ex.getMessage}"
+          None
+        }
```
ptravers: again don't need braces here but this is a personal style preference

money-java-servlet/src/main/scala/com/comcast/money/java/servlet/TraceFilter.scala
```diff
@@ -16,52 +16,94 @@
...
       }
     }
 
-    chain.doFilter(request, response)
+    Option(httpRequest.getHeader(B3TraceId)) match {
+      case Some(x) => parseHeaders(x)
+      case _ => None
+    }
```
ptravers: is this really a flatMap?

money-java-servlet/src/main/scala/com/comcast/money/java/servlet/TraceFilter.scala
```diff
@@ -16,52 +16,94 @@
...
-    incomingTraceId.foreach { traceId =>
-      response match {
-        case http: HttpServletResponse =>
-          http.addHeader(MoneyTraceHeader, traceId)
-        case _ =>
-          logger.warn("Unable to set money trace header on response, response type is not an HttpServletResponse ")
+    Option(httpRequest.getHeader(MoneyTraceHeader)) match {
+      case Some(x) => parseHeader(x)
+      case _ => None
+    }
```
ptravers: This should be a `flatMap`

money-java-servlet/src/main/scala/com/comcast/money/java/servlet/TraceFilter.scala
ptravers: I still feel a little bit that the parsers belong in Formatters but I am happy to defer to @paulcleary on that

ptravers:
Yeah generally pretty happy there's some alterations I want to convert the pattern matches in to flatMap and there's some excess braces (obviously not major). Only major question I have is if the parsers really belong in Formatters? But yeah aside from that happy to approve with minor refactors

money-java-servlet/src/main/scala/com/comcast/money/java/servlet/TraceFilter.scala
```diff
@@ -16,52 +16,94 @@
...
+  }
+
+  def parseB3TraceHeader(httpRequest: HttpServletRequest): Option[SpanId] = {
+
+    def parseHeaders(traceIdVal: String): Option[SpanId] = {
+      val maybeParentSpanId = Option(httpRequest.getHeader(B3ParentSpanId))
+      val maybeSpanId = Option(httpRequest.getHeader(B3SpanId))
```
ptravers: these should be `maybeB3SpanId` and `maybeB3ParentId`

money-java-servlet/src/main/scala/com/comcast/money/java/servlet/TraceFilter.scala
```diff
@@ -16,52 +16,94 @@
...
+      fromB3HttpHeaders(traceIdVal, maybeParentSpanId, maybeSpanId) match {
+        case Success(spanId) => Some(spanId)
+        case Failure(ex) => {
```
ptravers: I would prefer `b3SpanId`

money-java-servlet/src/main/scala/com/comcast/money/java/servlet/TraceFilter.scala
```diff
@@ -16,52 +16,94 @@
...
+      case _ => None
+    }
+  }
+
+  def parseB3TraceHeader(httpRequest: HttpServletRequest): Option[SpanId] = {
+
+    def parseHeaders(traceIdVal: String): Option[SpanId] = {
+      val maybeParentSpanId = Option(httpRequest.getHeader(B3ParentSpanId))
+      val maybeSpanId = Option(httpRequest.getHeader(B3SpanId))
+      fromB3HttpHeaders(traceIdVal, maybeParentSpanId, maybeSpanId) match {
```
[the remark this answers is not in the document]
MKnowl001: fromB3HttpHeaders is in core, which currently has no dependency on javaxServlet. javaxServlet has HttpServletRequest defined, so core would need a dependency on javaxServlet...which does not seem right.

fromB3HttpHeaders is in core because money-http-client also has a dependency on it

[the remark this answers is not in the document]
MKnowl001: More correctly money-http-client has a dependency on  Formatters, the containing object.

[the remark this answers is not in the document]
ptravers: Haha yeah. I see that now. Good call.

ptravers:
Looks good thanks for the contribution!

ptravers:
Looks good thanks for the contribution!

MKnowl001:
Thanks for all the great feedback.

MKnowl001:
Thanks for all the great feedback.

ptravers:
Cheers Mark have a great weekend!

PR #100, opened 2018-05-21
Add Akka Http Clients

ptravers:
One Client for insertion in to a stream

Another for Future based APIs


pauljamescleary:
@ptravers are we doing anything with this?  I want to work toward releasing 0.9.0.  If it is not needed perhaps we close?

ptravers:
Let me fix it and get it ready. I would love to get this out for 0.9

Issue #102, opened 2018-05-24
All X-B3 style headers should be hex encoded


MKnowl001:
Currently the X-B3-TraceId is "GUID" formatted, and the X-B3-ParentSpanId and X-B3-SpanId headers are decimal encoded. They should all be hex. We should expect hex strings when reading these headers, and output hex strings when creating them. https://github.com/openzipkin/b3-propagation

Existing: 
X-B3-TraceId: de305d54-75b4-431b-adb2-eb6b9e546013
X-B3-ParentSpanId: parent-id=3285573610483682037
X-B3-SpanId: span-id=3285573610483682037

Should be 
X-B3-TraceId: de305d5475b4431badb2eb6b9e546013
X-B3-ParentSpanId: 2d98b3ca80470af5
X-B3-SpanId: 2d98b3ca80470af5

PR #103, opened 2018-06-07
Fix issue #102, support X-B3 style headers correctly

MKnowl001:


money-core/src/main/scala/com/comcast/money/core/Formatters.scala
```diff
@@ -21,6 +21,21 @@ import com.comcast.money.api.SpanId
 import scala.util.Try
 
 object Formatters {
+  implicit class StringHexHelpers(s: String) {
```
ptravers: This may want to be private and I would prefer a name like `StringWithB3HexConversion` or something less generic than helpers.

the below shows some naming conventions
https://docs.scala-lang.org/overviews/core/implicit-classes.html

money-core/src/main/scala/com/comcast/money/core/Formatters.scala
```diff
@@ -33,14 +48,25 @@ object Formatters {
...
+  def toHttpHeader(spanId: SpanId): String = HttpHeaderFormat.format(spanId.traceId, spanId.parentId, spanId.selfId)
 
   def fromB3HttpHeaders(traceId: String, maybeParentSpanId: Option[String], maybeSpanId: Option[String]) = Try {
     (maybeParentSpanId, maybeSpanId) match {
-      case (Some(ps), Some(s)) => new SpanId(traceId, ps.toLong, s.toLong)
-      case (Some(ps), _) => new SpanId(traceId, ps.toLong)
-      case _ => new SpanId(traceId)
+      case (Some(ps), Some(s)) => new SpanId(traceId.toGuid, ps.fromHexStringToLong, s.fromHexStringToLong)
+      case (Some(ps), _) => new SpanId(traceId.toGuid, ps.fromHexStringToLong)
+      case (_, Some(s)) => new SpanId(traceId.toGuid, s.fromHexStringToLong, s.fromHexStringToLong) // root span
```
ptravers: rather than comment what do you think about just making these functions inside of fromB3Headers?

money-http-client/src/test/scala/com/comcast/money/http/client/TraceFriendlyHttpClientSpec.scala
```diff
@@ -68,9 +68,9 @@ class TraceFriendlyHttpClientSpec extends WordSpec with SpecHelpers
     verify(underTest.tracer).stopTimer(HttpTraceConfig.HttpResponseTimeTraceKey)
     verify(underTest.tracer).record("http-response-code", 200L)
     verify(httpUriRequest).setHeader("X-MoneyTrace", CoreSpanId.toHttpHeader(spanId))
-    verify(httpUriRequest).setHeader("X-B3-TraceId", spanId.traceId())
-    verify(httpUriRequest).setHeader("X-B3-ParentSpanId", spanId.parentId().toString)
-    verify(httpUriRequest).setHeader("X-B3-SpanId", spanId.selfId().toString)
+    verify(httpUriRequest).setHeader("X-B3-TraceId", spanId.traceId.replace("-", ""))
+    //    verify(httpUriRequest).setHeader("X-B3-ParentSpanId", spanId.parentId().t oHexString) // Does not get set because this is a root span
```
ptravers: remove this line please

money-core/src/main/scala/com/comcast/money/core/Formatters.scala
```diff
@@ -21,6 +21,21 @@ import com.comcast.money.api.SpanId
 import scala.util.Try
 
 object Formatters {
+  implicit class StringHexHelpers(s: String) {
+    def fromHexStringToLong: Long = java.lang.Long.parseUnsignedLong(s, 16)
```
ptravers: I think this should be a `Try` or `Option` since it is not safe

money-core/src/main/scala/com/comcast/money/core/Formatters.scala
```diff
@@ -33,14 +48,25 @@ object Formatters {
...
-      case (Some(ps), _) => new SpanId(traceId, ps.toLong)
-      case _ => new SpanId(traceId)
+      case (Some(ps), Some(s)) => new SpanId(traceId.toGuid, ps.fromHexStringToLong, s.fromHexStringToLong)
+      case (Some(ps), _) => new SpanId(traceId.toGuid, ps.fromHexStringToLong)
+      case (_, Some(s)) => new SpanId(traceId.toGuid, s.fromHexStringToLong, s.fromHexStringToLong) // root span
+      case _ => new SpanId(traceId.toGuid)
+    }
+  }
+
+  def toB3Headers(spanId: SpanId)(ft: String => Unit, fps: String => Unit, fs: String => Unit): Unit = {
```
ptravers: I'm not sure I understand the reasoning behind three separate methods. I would prefer the below

```
def toB3Headers(spanId: SpanId, addHeader: (String, String) => Unit): Unit = {
    val formatGuid = {
      // X-B3 style traceId's can be 8 octets long
      val traceIdHex = spanId.traceId.fromGuid
      if (traceIdHex.endsWith("0" * 16)) traceIdHex.substring(0, 16)
      else traceIdHex
    }
    addHeader("X-B3-TraceId", formatGuid)
    // No X-b3 parent if this is a root span
    if (spanId.parentId != spanId.selfId) addHeader("X-B3-ParentSpanId", spanId.parentId.toHexString)
    addHeader("X-B3-SpanId", spanId.selfId.toHexString)
  }
```

money-spring3/src/main/scala/com/comcast/money/spring3/MoneyClientHttpRequestInterceptor.scala
```diff
@@ -69,9 +69,11 @@ class MoneyClientHttpRequestInterceptor extends ClientHttpRequestInterceptor {
     SpanLocal.current foreach { span =>
       val headers = httpRequest.getHeaders
       headers.add("X-MoneyTrace", Formatters.toHttpHeader(span.info.id))
-      headers.add("X-B3-TraceId", span.info.id.traceId)
-      headers.add("X-B3-ParentSpanId", span.info.id.parentId.toString)
-      headers.add("X-B3-SpanId", span.info.id.selfId.toString)
+      Formatters.toB3Headers(span.info.id)(
```
ptravers: Since this function gives no control over what's set I would prefer the knowledge be kept in the Formatters I think this will prevent other locations in the library setting different header names.

`Formatters.toB3Headers(span.info.id, headers.add)`

MKnowl001: Agree, but we should also change fromB3Header and from/toHttpHeader. (Which should probably be renamed to from/toMoneyHeader)

MKnowl001:
I've refactored so that the header names are isolated in the Formatters object, and anyone who needs to read/write headers call fromHttpHeaders/toHttpHeaders. Formatters should probably be renamed.

Scala check is being used for the tests

"Can we keep / regurgitate the B3 sampling / flags?". I have not implemented this yet, but I think the code is in a better pace to do this. Probably need to stash them in the SpanId object during the fromHttpHeaders method, propagate from parent span to child, and  regurge in the toHttpHeaders.

money-core/src/test/scala/com/comcast/money/core/FormattersSpec.scala
```diff
@@ -16,40 +16,128 @@
...
-    "convert from a money  http header" in {
-      val spanId = new SpanId()
-      val test = Formatters.toHttpHeader(spanId)
-      Formatters.fromHttpHeader(test).get shouldBe spanId
+    "read a money  http header" in {
+      forAll { (traceIdValue: UUID, parentSpanIdValue: Long, spanIdValue: Long) =>
+        val expectedSpanId = new SpanId(traceIdValue.toString, parentSpanIdValue, spanIdValue)
+        val spanId = fromMoneyHeader(header => header match {
+          case MoneyTraceHeader => MoneyHeaderFormat.format(expectedSpanId.traceId, expectedSpanId.parentId, expectedSpanId.selfId)
+        }, _ => {})
```
ptravers: I would prefer a structure more like this without the default declared and not using the match block. A partial function will serve the same purpose and it's less verbose:
````    
"fail to read a badly formatted money  http header" in {
      forAll { (traceIdValue: String, parentSpanIdValue: String, spanIdValue: String) =>
        val spanId =
          fromMoneyHeader(
            getHeader = {
            case MoneyTraceHeader => MoneyHeaderFormat.format(traceIdValue, parentSpanIdValue, spanIdValue)
          }
          )

        spanId shouldBe None
      }
    }
```

money-core/src/test/scala/com/comcast/money/core/FormattersSpec.scala
```diff
@@ -16,40 +16,128 @@
...
+        val spanId = fromB3HttpHeaders(header => header match {
+          case B3TraceIdHeader => traceIdValue.toString.fromGuid
+          case B3ParentSpanIdHeader => parentSpanIdValue.toHexString
+          case B3SpanIdHeader => spanIdValue.toHexString
+        }, _ => {})
+        spanId shouldBe Some(expectedSpanId)
+        val maybeRootSpanId = fromB3HttpHeaders(header => header match {
+          case B3TraceIdHeader => traceIdValue.toString.fromGuid
+          case B3SpanIdHeader => spanIdValue.toHexString
+          case _ => null
```
ptravers: same idea partial function not restating the default log function

PR #110, opened 2018-10-20
Fix akka money

ptravers:
I haven't been able to get the timeout issue to appear but I fixed some ugly stuff with multiple actor systems being built per test suite

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -17,7 +17,7 @@ lazy val money = Project("money", file("."))
   publishLocal := {},
   publish := {}
 )
-.aggregate(moneyApi, moneyCore, moneyAspectj, moneyHttpClient, moneyJavaServlet, moneyWire, moneyKafka, moneySpring, moneySpring3)
+.aggregate(moneyApi, moneyAkka, moneyCore, moneyAspectj, moneyHttpClient, moneyJavaServlet, moneyWire, moneyKafka, moneySpring, moneySpring3)
 
 lazy val moneyApi =
   Project("money-api", file("./money-api"))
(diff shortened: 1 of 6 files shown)


pauljamescleary:
Thanks @ptravers,  I am seeing timeout errors.  I even adjusted the timeout to 10 seconds.

If travis builds it, I presume it is ok.  Really wish I could find out what is wrong locally as I have no way to independently verify before running a release to sonatype.  :/

```
Failed to stop [MoneyAkkaScope] within [10 seconds] 
-> / LocalActorRefProvider$$anon$1 null status=1 terminated
```

```
[info]   java.util.concurrent.TimeoutException: Futures timed out after [10 seconds]
[info]   at scala.concurrent.impl.Promise$DefaultPromise.ready(Promise.scala:255)
[info]   at scala.concurrent.impl.Promise$DefaultPromise.result(Promise.scala:259)
[info]   at scala.concurrent.Await$.$anonfun$result$1(package.scala:215)
[info]   at scala.concurrent.BlockContext$DefaultBlockContext$.blockOn(BlockContext.scala:53)
[info]   at scala.concurrent.Await$.result(package.scala:142)
[info]   at com.comcast.money.akka.Blocking$RichFuture.get(Blocking.scala:25)
[info]   at com.comcast.money.akka.acceptance.stream.TracedFlowSpec.$anonfun$new$2(TracedFlowSpec.scala:43)
```

pauljamescleary:
Interesting, I increased the timeout to 30 seconds, and the first spec in `TracedFlowSpec` failed the same, but the second test failed differently...

```
- MoneyExtension should pass a span through an asynchronous Akka Stream *** FAILED *** (34 milliseconds)
[info]   Names: Some(List(flow-3, flow-2, flow-1, flow-3, flow-2, flow-1)) were not Some(List(flow-3, flow-2, flow-1)) (TracedFlowSpec.scala:45)
```

pauljamescleary:
🎉 

@ptravers can you just modify `Blocking` and increase the timeout to `60.seconds`?  That seemed to work for me.

I suspect something is wrong on my local setup.

That allowed all tests to pass.

Once that is in, I will approve!

ptravers:
Huh it's almost like the test suite isn't coming up properly? Yeah will add that now! I'm glad something is making it work! that second error seems to be due to the prior test running adding it's spans to the CollectingSpanHandler and the BeforeEach not cleaning it out before the next test runs... really weird.

PR #114, opened 2018-11-30
update sample proj dependencies and migrate spring3 files

jonesleonard:
I updated the Spring and Jackson dependencies within the pom.xml file in the sample project to address the Spring and Jackson security issues. 

I also removed the money-spring3 module and moved the files to the money-spring module. 

There are some lingering issues. 

1. There are references to a `com.comcast.money.japi.TraceFriendlyExecutors` class in the [AppConfig class](https://github.com/Comcast/money/blob/master/samples/samples-springmvc/src/main/java/com/comcast/money/samples/springmvc/config/AppConfig.java#L10). 
2. I've run into some compiling issues with the move and attached a screenshot. Not sure if the issues are with my local setup or there are more changes I need to make/revert.


jonesleonard:
screen shot 2018-11-30 at 4 54 10 pm

PR #120, opened 2019-01-14
Add JMoney API with additional TWR and lambda helper APIs

HaloFour:
Adds the JMoney Java API bridge back to Money 0.9 with additional helper APIs supporting try-with-resources and lambdas.


money-core/src/main/scala/com/comcast/money/core/japi/JMoney.java
```diff
@@ -0,0 +1,577 @@
...
+
+import com.comcast.money.api.Note;
+import com.comcast.money.core.Money;
+import com.comcast.money.core.Tracer;
+
+/**
+ * Java API for working directly with money from Java code
+ */
+public class JMoney {
+    private static Tracer tracer;
```
pauljamescleary: this is never set (unless I missed it)?

Should always use `Money.Environment().tracer()`

money-core/src/main/scala/com/comcast/money/core/japi/JMoney.java
```diff
@@ -0,0 +1,577 @@
...
+    }
+
+    /**
+     * @return the {@link com.comcast.money.core.Tracer} to use for tracing
+     */
+    public static Tracer tracer() {
+
+        Tracer current = tracer;
+        if (current == null) {
+            return Money.Environment().tracer();
```
pauljamescleary: I would always return `Money.Environment().tracer()`, this way the `tracer` method becomes convenience.

money-core/src/test/scala/com/comcast/money/core/japi/JMoneySpec.scala
```diff
@@ -0,0 +1,361 @@
...
+
+      val captor = ArgumentCaptor.forClass(classOf[Note[_]])
+      verify(tracer).record(captor.capture())
+      val note = captor.getValue
+      val nil: AnyRef = null
+      note.name shouldBe "nill"
+      note.value shouldBe nil
+    }
+  }
+}
```
pauljamescleary: ahh, new line!

money-core/src/main/scala/com/comcast/money/core/japi/JMoney.java
HaloFour: I'm only allowing `Tracer` to be set in order to support the unit tests.  Do you have a preferred pattern to follow?  This one is similar to what `SpringMoney` does.

money-core/src/main/scala/com/comcast/money/core/japi/JMoney.java
HaloFour: There is a `protected void setTracer` method, used by the unit tests to pass a mock `Tracer`

HaloFour:
I've also been toying with an async version of the API which would pop the span and only stop it when the returned future completes.  It involves a bit more plumbing and would want to review the concept before I'd suggest adding it, perhaps in a future PR.

Here's a gist of the implementation: https://gist.github.com/HaloFour/8c55faef7135355ae31a3ebabf2dd9e3

money-core/src/main/scala/com/comcast/money/core/japi/JMoney.java
pauljamescleary: Is it possible to allow unit tests the ability to set the static tracer, and just initialize it by default to `Money.Environment().tracer()`?

PR #124, opened 2019-03-20
Use List instead of Stack for thread local spans

HaloFour:
[`scala.collection.mutable.Stack`](https://www.scala-lang.org/api/2.12.0/scala/collection/mutable/Stack.html) is deprecated as of Scala 2.12 and the recommendation is to just use a [`scala.collection.immutable.List`](https://www.scala-lang.org/api/2.12.0/scala/collection/immutable/List.html) instead.

### DIFF
--- a/money-core/src/main/scala/com/comcast/money/core/internal/SpanLocal.scala
+++ b/money-core/src/main/scala/com/comcast/money/core/internal/SpanLocal.scala
@@ -23,11 +23,11 @@ import scala.collection.mutable.Stack
 trait SpanContext {
   def push(span: Span): Unit
 
-  def pop: Option[Span]
+  def pop(): Option[Span]
 
   def current: Option[Span]
 
-  def clear: Unit
+  def clear(): Unit
 }
 
 /**
@@ -37,7 +37,7 @@ trait SpanContext {
 object SpanLocal extends SpanContext {
 
   // A stack of span ids for the current thread
-  private[this] val threadLocalCtx = new ThreadLocal[Stack[Span]]
+  private[this] val threadLocalCtx = new ThreadLocal[List[Span]]
 
   private lazy val mdcSupport = new MDCSupport()
 
@@ -46,36 +46,33 @@ object SpanLocal extends SpanContext {
   override def current: Option[Span] = Option(threadLocalCtx.get).flatMap(_.headOption)
 
   override def push(span: Span): Unit =
-    if (span != null)
+    if (span != null) {
       Option(threadLocalCtx.get) match {
-        case Some(stack) =>
-          stack.push(span)
-          setSpanMDC(Some(span.info.id))
-          setSpanNameMDC(Some(span.info.name))
+        case Some(current) =>
+          threadLocalCtx.set(span :: current)
         case None =>
-          threadLocalCtx.set(new Stack[Span]())
-          push(span)
+          threadLocalCtx.set(List(span))
       }
+      setSpanMDC(Some(span.info.id))
+      setSpanNameMDC(Some(span.info.name))
+    }
 
   override def pop(): Option[Span] =
-    Option(threadLocalCtx.get).map {
-      stack =>
-        // remove the current span in scope for this thread
-        val spanId = stack.pop()
-
-        // rolls back the mdc span to the parent if present, if none then it will be cleared
-        val head = stack.headOption
+    threadLocalCtx.get match {
+      case span :: rest =>
+        threadLocalCtx.set(rest)
+        val head = rest.headOption
         setSpanMDC(head.map(_.info.id))
         setSpanNameMDC(head.map(_.info.name))
-
-        spanId
+        Some(span)
+      case _ => None
     }
 
   /**
    * Clears the entire call stack for the thread
    */
-  override def clear() = {
-    if (threadLocalCtx != null) threadLocalCtx.remove()
+  override def clear(): Unit = {
+    threadLocalCtx.remove()
 
     setSpanMDC(None)
     setSpanNameMDC(None)
--- a/money-core/src/main/scala/com/comcast/money/core/internal/SpanLocal.scala
+++ b/money-core/src/main/scala/com/comcast/money/core/internal/SpanLocal.scala
@@ -18,8 +18,6 @@ package com.comcast.money.core.internal
 
 import com.comcast.money.api.Span
 
-import scala.collection.mutable.Stack
-
 trait SpanContext {
   def push(span: Span): Unit
 
@@ -37,7 +35,7 @@ trait SpanContext {
 object SpanLocal extends SpanContext {
 
   // A stack of span ids for the current thread
-  private[this] val threadLocalCtx = new ThreadLocal[List[Span]]
+  private[this] val threadLocalCtx = ThreadLocal.withInitial[List[Span]](() => Nil)
 
   private lazy val mdcSupport = new MDCSupport()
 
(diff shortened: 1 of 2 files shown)


money-core/src/main/scala/com/comcast/money/core/internal/SpanLocal.scala
```diff
@@ -46,36 +46,33 @@ object SpanLocal extends SpanContext {
...
+      case _ => None
     }
 
   /**
    * Clears the entire call stack for the thread
    */
-  override def clear() = {
-    if (threadLocalCtx != null) threadLocalCtx.remove()
+  override def clear(): Unit = {
+    threadLocalCtx.remove()
```
pauljamescleary: unclear on why we removed the threadLocalCtx != null check.  it seemingly doesn't make sense that it is there, but not remembering why it was there in the first place leads me to believe we should just leave it.

HaloFour: I'm not sure under what circumstance that field could be `null`.  Maybe the intention was to check if `threadLocalCtx.get` was `null`?

I can add the `null` check back if it makes you feel more comfortable, but that begs the question as to why we're not checking that in the other methods.

money-core/src/main/scala/com/comcast/money/core/internal/SpanLocal.scala
```diff
@@ -46,36 +46,33 @@ object SpanLocal extends SpanContext {
...
-        val head = stack.headOption
+    threadLocalCtx.get match {
+      case span :: rest =>
+        threadLocalCtx.set(rest)
+        val head = rest.headOption
         setSpanMDC(head.map(_.info.id))
         setSpanNameMDC(head.map(_.info.name))
-
-        spanId
+        Some(span)
```
pauljamescleary: probably could improve this slightly...

```scala
rest.headOption.foreach { head => 
  setSpanMDC(head.info.id)
  setSpanNameMDC(head.info.name)
}
Some(span)
```

pauljamescleary:
I don't know why codecov is complaining so much, but I think we are ok

money-core/src/main/scala/com/comcast/money/core/internal/SpanLocal.scala
HaloFour: The only issue with this is that I'd still want to call those `setSpan*` methods with `None` if there is no `Span` so that the MDC keys are cleared.

money-core/src/main/scala/com/comcast/money/core/internal/SpanLocal.scala
```diff
@@ -37,7 +37,7 @@ trait SpanContext {
 object SpanLocal extends SpanContext {
 
   // A stack of span ids for the current thread
-  private[this] val threadLocalCtx = new ThreadLocal[Stack[Span]]
+  private[this] val threadLocalCtx = new ThreadLocal[List[Span]]
```
pauljamescleary: I think we want this mutable, using ArrayBuffer

money-core/src/main/scala/com/comcast/money/core/internal/SpanLocal.scala
```diff
@@ -46,35 +46,32 @@ object SpanLocal extends SpanContext {
...
 
   override def pop(): Option[Span] =
-    Option(threadLocalCtx.get).map {
-      stack =>
-        // remove the current span in scope for this thread
-        val spanId = stack.pop()
-
-        // rolls back the mdc span to the parent if present, if none then it will be cleared
-        val head = stack.headOption
+    threadLocalCtx.get match {
```
pauljamescleary: interesting that you dropped the `Option` here, I think technically it is ok but I don't like pattern matching on a null

money-core/src/main/scala/com/comcast/money/core/internal/SpanLocal.scala
HaloFour: The `ThreadLocal` itself is mutable, I'm replacing its value with the updated `List[Span]` on each push/pop.  The net effect is the same.  But if you'd prefer a mutable collection I can change the implementation.

ptravers: https://docs.scala-lang.org/overviews/collections/performance-characteristics.html

ptravers: ListBuffer will have the same performance characteristics as list though

https://www.scala-lang.org/api/current/scala/collection/mutable/ListBuffer.html

ptravers: after talking with @HaloFour  I don't think we gain anything from using a mutable collection. I mostly feel like analyzing performance without metrics won't glean us a whole lot either.

money-core/src/main/scala/com/comcast/money/core/internal/SpanLocal.scala
```diff
@@ -37,7 +35,7 @@ trait SpanContext {
 object SpanLocal extends SpanContext {
 
   // A stack of span ids for the current thread
-  private[this] val threadLocalCtx = new ThreadLocal[Stack[Span]]
+  private[this] val threadLocalCtx = ThreadLocal.withInitial[List[Span]](() => Nil)
```
pauljamescleary: cool!

money-core/src/main/scala/com/comcast/money/core/internal/SpanLocal.scala
pauljamescleary: @ptravers I agree, using a linked list is good from an allocation perspective

money-core/src/main/scala/com/comcast/money/core/internal/SpanLocal.scala
pauljamescleary: I am ok leaving this as is, I did some reading and testing and it seems like what you had was ok.  Having an initial value always there is cool

HaloFour:
My last commit included a breaking change to `MDCSupport`.  Is that considered a part of the public API or more an internal utility class?

Speaking of breaking changes, I'd also like to push a commit to fix the `propogate` spelling error which is fairly pervasive.  I believe that there is only a single method in `MDCSupport` which contains that misspelling, but it's also used numerous times as a parameter name which could be a source breaking change for Scala programs using named arguments.  I wanted to gauge how everyone felt about such breaking changes.

money-core/src/main/scala/com/comcast/money/core/internal/MDCSupport.scala
```diff
@@ -38,10 +38,14 @@ class MDCSupport(enabled: Boolean = Money.Environment.enabled) {
   private val MoneyTraceKey = "moneyTrace"
   private val SpanNameKey = "spanName"
 
-  def setSpanMDC(spanId: Option[SpanId]): Unit = if (enabled) {
-    spanId match {
-      case Some(id) => MDC.put(MoneyTraceKey, MDCSupport.format(id))
-      case None => MDC.remove(MoneyTraceKey)
+  def setSpanInfoMDC(spanInfo: Option[SpanInfo]): Unit = if (enabled) {
```
ptravers: @HaloFour  I'm assuming this is the breaking change you are referring too?

I would say that anything in core/internal should not be being accessed by outside users. Not sure how we would enforce privacy here.

HaloFour: Yes.  And I can keep the other methods around, mark them as deprecated if we feel that's appropriate.  To be honest I used the change to bring up the bigger issue of the `propogate` misspelling which impacts one method name as well as numerous parameter names.

HaloFour: I've decided to back off of those changes for now.  There are numerous ways that I think the ergonomics of the API can be improved but that can probably wait until a major release.

ptravers: @HaloFour I thought about this again recently and realized I was totally wrong. We should just make this package private to Money i.e. `private[money]`.

I will add a issue to our stack to make everything in internal package private to Money unless @paulcleary or yourself have any objections?

HaloFour:
Addressed by other PRs

PR #131, opened 2020-09-30
Add traceparent header and hex formatting

HaloFour:
Adds reading/writing of W3C Trace Context `traceparent` header.  Also adds configuration option to format trace ID and span ID in MDC as hex in the same format as Trace Context IDs.  Fixes formatting of ZipKin B3 headers.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -185,7 +185,7 @@ def aspectjProjectSettings = projectSettings ++ Seq(
 
 def basicSettings =  Defaults.itSettings ++ SbtScalariform.scalariformSettings ++ Seq(
   organization := "com.comcast.money",
-  version := "0.9.1-SNAPSHOT",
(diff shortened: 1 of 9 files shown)


ptravers:
Documentation for the trace header https://www.w3.org/TR/trace-context/

HaloFour:
Travis is broken which is very annoying.

Thanks for posting the docs, @ptravers.

Since I also fixed the Zipkin B3 headers here are those docs:  https://github.com/openzipkin/b3-propagation

The existing Money implementation does not pad out the span IDs to 16 hex characters as is required so I have fixed that.

ptravers:
I will try to get travis working but may refactor to use github actions

ptravers:
Looks like an sbt issue. still triaging

HaloFour:
Clearly not ready at this time.

PR #134, opened 2020-10-01
Add W3C Trace Context headers and OpenTelemetry-compat ID formatting

HaloFour:
Adds reading/writing of W3C Trace Context `traceparent` header. Also adds configuration option to format trace ID and span ID in MDC as hex in the same format as OpenTelemetry IDs. Fixes formatting of ZipKin B3 headers which need to be padded to 16 characters.


pauljamescleary (approved):
Looks great @HaloFour

pauljamescleary:
LGTM!

PR #136, opened 2020-10-02
Update to latest Scala 2.12 and 2.13 versions, update ScalaTest

HaloFour:


pauljamescleary (approved):
Great!  However, seems like something with scoverage is off, as http-client has a code coverage of 30% all of a sudden.  Once the build is passing LGTM

HaloFour:
Still working on those details.  The change to `junit-interface` was causing the JUnit tests to not run.  I _think_ that's fixed now but I'm trying to resolve an issue with the tests in `money-kafka`.  One of the tests depended on being able to obtain the configuration from Kafka after creating the producer and that's no longer possible.

PR #137, opened 2020-10-02
Clean up some warnings and a typo in a public method

HaloFour:
Clean up most of the Scala 2.12 warnings.  Many of the Scala 2.13 warnings cannot currently be addressed with code that can cross-compile with Scala 2.12.

Replace use of deprecated `Stack[Span]` in `SpanLocal` with `List[Span]` treated as a stack.

Fix typo in name of `MDCSupport.propagateMDC` method.

### DIFF
--- a/money-akka/src/main/scala/com/comcast/money/akka/SpanContextWithStack.scala
+++ b/money-akka/src/main/scala/com/comcast/money/akka/SpanContextWithStack.scala
@@ -93,7 +93,7 @@ class SpanContextWithStack() extends SpanContext {
    * @return Option[Span]
    */
 
-  override def pop: Option[Span] =
+  override def pop(): Option[Span] =
     stack.headOption match {
       case someSpan: Some[Span] =>
         stack = stack.tail
@@ -112,7 +112,7 @@ class SpanContextWithStack() extends SpanContext {
    * @return
    */
 
-  override def current: Option[Span] = stack.headOption
+  override def current(): Option[Span] = stack.headOption
 
   /**
    * CAUTION THIS WILL CAUSE ALL THE STARTED SPANS TO NOT BE STOPPED
@@ -122,5 +122,5 @@ class SpanContextWithStack() extends SpanContext {
    * empties the Stack of all spans
    */
 
-  override def clear: Unit = stack = List.empty[Span]
+  override def clear(): Unit = stack = List.empty[Span]
 }
--- a/money-akka/src/main/scala/com/comcast/money/akka/SpanContextWithStack.scala
+++ b/money-akka/src/main/scala/com/comcast/money/akka/SpanContextWithStack.scala
@@ -112,7 +112,7 @@ class SpanContextWithStack() extends SpanContext {
    * @return
    */
 
-  override def current(): Option[Span] = stack.headOption
+  override def current: Option[Span] = stack.headOption
 
   /**
    * CAUTION THIS WILL CAUSE ALL THE STARTED SPANS TO NOT BE STOPPED
--- a/money-core/src/main/scala/com/comcast/money/core/internal/MDCSupport.scala
+++ b/money-core/src/main/scala/com/comcast/money/core/internal/MDCSupport.scala
@@ -18,7 +18,7 @@ package com.comcast.money.core.internal
 
 import java.util.Map
 
-import com.comcast.money.api.SpanId
+import com.comcast.money.api.{ Span, SpanId }
 import com.comcast.money.core.Money
 import org.slf4j.MDC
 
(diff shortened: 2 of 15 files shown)


HaloFour:
The difference between empty parameter lists between `SpanContext` and `SpanLocal` generates a bunch of warnings, but I'm finding that making changes to either breaks existing code.  I did make a few changes in less impacted areas but if that's controversial I can walk it back.  If these changes are leading towards a potential `1.0` release then perhaps it's worth taking the breaking changes at this time to make the API consistent.

HaloFour:
Was reading up on idiomatic Scala and it sounds like empty parameter lists should be omitted when the method has no side effects.  Given this my opinion would be that we should change the `SpanContext` trait to be as follows:

```scala
trait SpanContext {
  def push(span: Span): Unit
  def pop(): Option[Span]
  def current: Option[Span]
  def clear(): Unit
}
```

As `pop` and `clear` are expected to have the side effect of modifying the ambient span.  I think `SpanLocal` should also be updated to follow the same convention.

ptravers:
I would prefer to get the W3C Trace Header updates out before making a breaking release. Happy to do a breaking release after that. I'm a little concerned about making a 1.0 release without going through and doing a deep clean. There's a bunch of stuff that needs to be ripped out of the akka package and I imagine a lot of the other packages are so decrepit that they should really be rejuvenated before we got to 1.0. At some point we do need to have a discussion about what we want to do with Money. I personally feel like converting Money to being a shim for OpenTelemetry (OT) and asking our users to make the switch to OT would be the most responsible solution for the library going forward.

cc: @paulcleary what are your thoughts?

money-core/src/main/scala/com/comcast/money/core/internal/MDCSupport.scala
```diff
@@ -50,27 +50,22 @@ class MDCSupport(
...
+        MDC.put(SpanNameKey, s.info.name)
+      case None =>
+        MDC.remove(MoneyTraceKey)
+        MDC.remove(MoneyTraceKey)
+        MDC.remove(SpanNameKey)
     }
   }
 
-  def propogateMDC(submittingThreadsContext: Option[Map[_, _]]): Unit = if (enabled) {
+  def propagateMDC(submittingThreadsContext: Option[Map[String, String]]): Unit = if (enabled) {
```
ptravers, line 65: Is this the only breaking change? Can we just instead deprecate the incorrect name and add a second method with the correct name?

I am actually more concerned about the type changes being made here and what effect that will have on integrated codebases. I also don't have any context as to why we need to wildcard the type here.

money-akka/src/main/scala/com/comcast/money/akka/SpanContextWithStack.scala
```diff
@@ -122,5 +122,5 @@ class SpanContextWithStack() extends SpanContext {
    * empties the Stack of all spans
    */
 
-  override def clear: Unit = stack = List.empty[Span]
+  override def clear(): Unit = stack = List.empty[Span]
```
ptravers, line 125: :+1:

money-core/src/main/scala/com/comcast/money/core/internal/MDCSupport.scala
HaloFour: I don't know if the changes to `SpanContext` are source-breaking or binary-breaking.  They might count also. 

Old versions of slf4j returned the raw `Map` type with the context map.

As for the name change, that's up for debate.  IMO if this is a change on the road to 1.x it's better to accept the breaking changes now, but we can also temporarily include the deprecated version until we do release that version.

HaloFour:
I'd also not want to rev Money to 1.0 until OpenTelemetry does, so that would be another reason to avoid (or mitigate) breaking changes.

ptravers:
Yeah that's a great point! I think everyone in this thread knows already but I created an issue related to this https://github.com/Comcast/money/issues/138

money-core/src/main/scala/com/comcast/money/core/internal/MDCSupport.scala
pauljamescleary: How is this breaking, just curious?  The method is internal to the money runtime isn't it?  I guess as long as anything that calls this method is also updated it is ok?

pauljamescleary (approved):

money-core/src/main/scala/com/comcast/money/core/internal/MDCSupport.scala
HaloFour: It's technically a public class and used by other modules in the project to propagate MDC across thread boundaries.  It's possible that an external client took a dependency on the class directly if they were tracing around some other asynchronous library not covered by one of Money's projects.

PR #140, opened 2020-10-07
Implement OpenTelemetry Tracing Interfaces and APIs

HaloFour:
Implement OpenTelemetry Tracing interfaces.


HaloFour:
I think I've completed the main work in getting Money to implement the OpenTelemetry APIs.  I've tried to keep as much of the Money interfaces intact as I could, although I did make some quality-of-life changes that I had been wanting from Money.  Some of the new OTel properties like Kind, Status and Description I append as notes to the span.  I track everything as nanos (using a drift-correcting monotonic clock) but convert to millis/micros for the SpanHandler.  All existing tests pass.  The build itself would pass if it weren't for the fact that code coverage has tanked.

Before I go further down this rabbit hole I'd let to get some feedback as to whether or not this is the correct route to take.

HaloFour:
TD;DR:

1. Money `Span` extends from OTel `Span`
2. Money `Tracer` extends from `MoneyTracer` which extends from OTel `Tracer`
    * Added the intermediate Java interface `MoneyTracer` to declare the members with covariant returns so that they return Money `Span` instead of OTel `Span`
3. `SpanLocal` populates `Context`
    * This feels kinda hacky since everything in `Context` is scope-driven.  Having to keep track of the `Context` instances in order to "pop" the current thread context feels "wrong".  Might be worth reevaluating `push`+`pop` pair with a method that returns a `Scope` instead.
4. Added `Event` to represent an OTel event which can be added to a `Span`.
    * The list of events is added to the `SpanInfo` which is passed to the `SpanHandler`
5. Expanded `Note` to support lists of `String`, `Long`, `Boolean` and `Double`
    * Added new factory methods to add each list, but this may be overkill since it's not necessary to support OTel 0.9.1
6. Added Money `Span.Builder` interface that extends OTel `Span.Builder` interface, again for covariant returns
    * Added `CoreSpanBuilder` which implements this interface and wraps the Tracer's `SpanFactory` to create and start new spans.
7. Added `Clock` trait with a `SystemClock` implementation with millisecond resolution and `NanoClock` monotonic implementation that uses the high-resolution timers with drift correction
    * This is based both on the interface in OTel SDK and the implementation currently being used in XVP
8. Some other stuff I'm probably forgetting at the moment.

money-api/src/main/java/com/comcast/money/api/MoneyTracer.java
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.comcast.money.api;
+
+import io.opentelemetry.context.Scope;
+import io.opentelemetry.trace.Tracer;
+
+public interface MoneyTracer extends Tracer {
```
ptravers, line 22: Is this for Java compatibility?

ptravers:
Took a first look over it nothing major. Looks fine will go over it again with a fine tooth comb.

money-api/src/main/java/com/comcast/money/api/MoneyTracer.java
HaloFour: It's for covariant returns, I wanted to be able to return the Money flavor of `Span` instead of the OTel `Span`, that way consumers wouldn't have to cast.  Doesn't seem like Scala supports covariant returns directly but Java lets me override the members in an interface that extends from the original interface.

money-core/src/main/scala/com/comcast/money/core/CoreSpanBuilder.scala
```diff
@@ -0,0 +1,100 @@
...
+  override def setAttribute(key: String, value: Long): Span.Builder = appendAction({ _.setAttribute(key, value) })
+
+  override def setAttribute(key: String, value: Double): Span.Builder = appendAction({ _.setAttribute(key, value) })
+
+  override def setAttribute(key: String, value: Boolean): Span.Builder = appendAction({ _.setAttribute(key, value) })
+
+  override def setAttribute[T](key: AttributeKey[T], value: T): Span.Builder = appendAction({ _.setAttribute(key, value) })
+
+  private def appendAction(f: Span => Unit): Span.Builder = {
+    actions = f :: actions
```
pauljamescleary, line 66: this is disingenuous as you are _prepending_ here ;P

money-core/src/main/scala/com/comcast/money/core/CoreSpanBuilder.scala
```diff
@@ -0,0 +1,100 @@
...
+
+  override def setStartTimestamp(startTimestamp: Long): Span.Builder = {
+    this.startTimeNanos = startTimestamp
+    this
+  }
+
+  override def startSpan(): Span = {
+    val newSpan = parentSpan match {
+      case Some(span) => spanFactory.childSpan(spanName, span, sticky)
+      case None => spanFactory.newSpan(spanName)
```
pauljamescleary, line 83: seems like we should have a `spanFactory.newSpan` that takes an `Option[ParentSpan]`

money-core/src/main/scala/com/comcast/money/core/CoreSpanBuilder.scala
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    actions.foreach { _(newSpan) }
```
pauljamescleary, line 86: not sure what an action is, but I surmise that this is like an event listener that gets fired at certain points

HaloFour: Kind of a hacky way to keep track of the mutations made with the SpanBuilder (adding attributes, etc.) that I want to then "playback" on the span once I've created it.  I'm sure there's a better way to do this.

money-core/src/main/scala/com/comcast/money/core/CoreSpanBuilder.scala
HaloFour: Not a bad idea.  I was following the existing conventions, for better or worse.

money-core/src/main/scala/com/comcast/money/core/CoreSpanBuilder.scala
HaloFour: I've gone with a `List[Note[_]]` instead.

HaloFour:
This might've taken on a life of its own.  The OpenTelemetry API interfaces are now weaved through much of money-core and money-api.  I tried to limit the breaking changes in money-api but I did opt for some quality-of-life changes.  We can discuss whether or not they are worthwhile.  There are more breaking changes in money-core, mostly to the parameters of the case classes.

Code coverage is back above the required minimum, but I still have some more tests to write to get good coverage over some of the changed classes.

Thoughts?

HaloFour:
Oh, also, bumped the version number to `0.10.9.0.1-SNAPSHOT`.  A bit of a mouthful but I wanted to rev the version to indicate possible breaking changes but I didn't want to shoot for `1.x` yet due to the OpenTelemetry API not being GA.  This version embeds the OpenTelemetry API version so that the two can evolve in lock-step until OpenTelemetry is released.

HaloFour:
I'm taking the WIP label off of this PR.  I _think_ I have the APIs, code coverage and docs in a good place.  This is a big PR and pretty far reaching due to the interfaces and members that have been added.

money-api/src/main/java/com/comcast/money/api/SpanFactory.java
```diff
@@ -33,7 +33,7 @@
      * @return a child span with trace id and parent id from trace context header or a new root span if the
      * traceContextHeader is malformed.
      */
-    Span newSpanFromHeader(String childName, Function1<String,String> getHeader);
+    Span newSpanFromHeader(String childName, Function<String, String> getHeader);
```
HaloFour, line 36: This is a binary breaking change but it enables both Java and Scala consumers to pass lambdas.

money-api/src/main/java/com/comcast/money/api/SpanInfo.java
```diff
@@ -26,32 +30,67 @@
...
+    /**
+     * @return a list of all of the events that were recorded on the span.
+     */
+    List<Event> events();
+
     /**
      * @return the time in milliseconds when this span was started
      */
-    Long startTimeMillis();
+    long startTimeMillis();
```
HaloFour, line 41: Changed these methods to return `long` as it avoids the allocation and it doesn't seem that these methods are ever intended to return `null`.

money-core/src/main/scala/com/comcast/money/core/CoreSpan.scala
```diff
@@ -32,65 +43,139 @@ import scala.collection.concurrent.TrieMap
  */
 case class CoreSpan(
   id: SpanId,
-  name: String,
+  var name: String,
```
HaloFour, line 46: This would be a breaking change for any consumer directly using `CoreSpan` instead of going through `SpanFactory`.  I don't think it's avoidable, though, and these classes should probably be considered internal.

money-core/src/main/scala/com/comcast/money/core/internal/SpanLocal.scala
```diff
@@ -34,42 +33,31 @@ trait SpanContext {
...
-
   private lazy val mdcSupport = new MDCSupport()
 
   import mdcSupport._
 
-  override def current: Option[Span] = Option(threadLocalCtx.get).flatMap(_.headOption)
+  override def current: Option[Span] = fromContext(Context.current)
 
-  override def push(span: Span): Unit =
+  override def push(span: Span): Scope =
```
HaloFour, line 42: The removal of `pop` would definitely be a breaking change for anyone who directly took a dependency on `SpanLocal` or `SpanContext`, which are internal APIs.

ptravers: Can we make those package private as part of this PR?

money-core/src/main/scala/com/comcast/money/core/CoreSpan.scala
ptravers: If we are making this mutable I would personally prefer to then make the CoreSpan package private

HaloFour: Seems like all of the `Core*` classes can be made package private without breaking the other projects with the exception of `CoreSpanInfo` which is used by a couple of other projects directly (e.g. `money-wire` and `money-spring`).  As `CoreSpanInfo` is really just a data carrier I wonder if it's worth moving the implementation to `money-api` with a factory method defined on `SpanInfo` in order to create it.

money-core/src/main/scala/com/comcast/money/core/internal/SpanLocal.scala
HaloFour: Making these package private might be tricky as a number of the other projects use `SpanLocal` directly and `SpanContext` is exposed directly from `Tracer`.  Although, with the exception of `clear`, all of the members should be accessible through the `Tracer` object instead.  I can look into it but it might be more involved.

ptravers: Okay. I wasn't sure if we could make it private to money and still have access in all the other packages

build.sbt
```diff
@@ -167,7 +170,7 @@ def aspectjProjectSettings = projectSettings ++ Seq(
 
 def basicSettings =  Defaults.itSettings ++ Seq(
   organization := "com.comcast.money",
-  version := "0.9.2-SNAPSHOT",
+  version := "0.10.9.0.1-SNAPSHOT",
```
pauljamescleary, line 173: what kind of version number is this XD

money-akka/src/main/scala/com/comcast/money/akka/SpanContextWithStack.scala
```diff
@@ -79,28 +81,10 @@ class SpanContextWithStack() extends SpanContext {
...
-    stack.headOption match {
-      case someSpan: Some[Span] =>
-        stack = stack.tail
-        someSpan
-
-      case None => None
-    }
+  override def push(span: Span): Scope = {
+    stack = span :: stack
+    () => stack = stack.drop(1)
```
pauljamescleary, line 86: can you explain what Scope is here?  Looks like we are mutating the thread local stack by removing head, I surmise this is `Unit` as well

money-api/src/main/java/com/comcast/money/api/Note.java
```diff
@@ -16,38 +16,56 @@
 
 package com.comcast.money.api;
 
+import java.util.Objects;
+
+import io.opentelemetry.common.AttributeKey;
+import io.opentelemetry.common.AttributeType;
+
```
pauljamescleary, line 23: Should (or can) `Note<T>` extend `Attribtute<T>` from open telementry?

money-core/src/main/scala/com/comcast/money/core/CoreSpan.scala
pauljamescleary: Hmm, seems like a leak if other projects rely on `CoreSpanInfo` instead of `SpanInfo` interface.  Perhaps I got lazy?

build.sbt
HaloFour: I wanted to indicate that it was Money `0.10.x` and that it was evolving with OpenTelemetry `0.9.1` but I botched that up and added an extra `0`.  I can remove that middle `0`, would you be more amenable to version `0.10.9.1`?

money-akka/src/main/scala/com/comcast/money/akka/SpanContextWithStack.scala
HaloFour: `Scope` is a functional interface equivalent to `() => ()`.  The `push` function adds the span to the head of the list and then returns the function to pop the head from that list.

pauljamescleary (commented):
This is awesome!  The test would be, how does the sample spring app behave with the changes?

HaloFour:
@ptravers Also suggested we could have a `SpanHandler` implementation that wraps the `SpanProcessor` interface from OpenTelemetry SDK and the reference implementation.  I think that would be awesome and would allow the use of OTel compatible exporters from Money, including ZipKin and Jaeger.  But that will be a separate PR.  😁

money-akka/src/main/scala/com/comcast/money/akka/SpanContextWithStack.scala
HaloFour: @ptravers I'm less familiar with the span state that the money-akka project needs to track, do you mind checking this out to make sure that it won't mess anything up?

HaloFour:
Output from sample SpringMVC project:

```
[STDOUT] 2020-10-13 22:37:14,138   Span: [ span-id=7555669772236629277 ][ trace-id=28a095a4-d376-42d5-9c75-74214b516263 ][ parent-id=7555669772236629277 ][ span-name=SAMPLE_CONTROLLER ][ app-name=SpringMVCSample ][ start-time=1602643033916 ][ span-duration=221771 ][ span-success=true ][ CONTROLLER_INPUT=world ]
[STDOUT] 2020-10-13 22:37:14,138   Span: [ span-id=2403311888510551783 ][ trace-id=28a095a4-d376-42d5-9c75-74214b516263 ][ parent-id=685214453742432929 ][ span-name=NESTED_SERVICE ][ app-name=SpringMVCSample ][ start-time=1602643034081 ][ span-duration=56264 ][ span-success=true ][ CONTROLLER_INPUT=world ][ nested-wait=300.0 ][ http-call-duration=54187827 ][ http-response-code=200 ]
[STDOUT] 2020-10-13 22:37:14,138   Span: [ span-id=685214453742432929 ][ trace-id=28a095a4-d376-42d5-9c75-74214b516263 ][ parent-id=7555669772236629277 ][ span-name=ROOT_SERVICE ][ app-name=SpringMVCSample ][ start-time=1602643033917 ][ span-duration=219773 ][ span-success=true ][ CONTROLLER_INPUT=world ][ ROOT_NAME=world ][ wait=112.0 ]
```

build.sbt
pauljamescleary: Ahh, ok, I didn't catch the linkage to open telemetry.

We would need to move both thought.  For example, if we have a patch, we would be `0.10.1, 0.10.2` even if open tel stays `0.9.1`.

Also, unclear on how much change is going on in open tel land.  Like, if they release 0.9.2 and there are no substantive changes, would we do a new release just to stay in line with the version numbers?

This also begs the critical question, we need to update our release process to go through GHA and use github secrets so that anyone can push the button (different topic)

HaloFour: The OTel Java project is currently versioning once a month.  They usually only rev the minor release, but this past month there was a problem and they had to cut an interim release, hence `0.9.1`.  This will require ongoing maintenance until OTel publishes a GA release and I am willing to keep up with any necessary changes.  These monthly releases have included breaking changes, which is why I thought it would be worthwhile to link the Money version to the OTel version to make it apparent how the versions are associated to each other.

I do agree that if Money needs to rev to keep up with the OTel API releases that simplifying/automating the release process would be a big help.

@ptravers You have been more involved with OTel than I have, what are your thoughts on their timeframe to GA?  I also asked in their [Gitter channel](https://gitter.im/open-telemetry/opentelemetry-java).

ptravers: They are currently racing to hit GA very soon. I would guess it's unlikely there are significant changes before GA. Java implementation is in clean up mode for GA

money-api/src/main/java/com/comcast/money/api/SpanFactory.java
ptravers: You're going to have to forgive me but what would the issues be with breaking binary compatibility here? Just trying to think when that would come up for a user in normal usage

HaloFour: I suspect that this API is probably rarely used directly.  It might be used if a consumer is using an HTTP framework that one of the existing Money projects doesn't support out of the box.  Currently calling this API from Java is kludgy given that Scala's `Function1[T, R]` type is not recognized as a functional interface, so Java users have to generate an anonymous class extending from `AbstractFunction<T, R>`.  This change enables both Java and Scala to call the function passing a lambda, since Scala 2.12+ supports functional interaces.

build.sbt
HaloFour: Any suggestions on the version?  I thought it would be useful to tie the Money version to the OTel version to make it obvious which version of Money you'd want to integrate with existing OTel libraries, especially as OTel is still evolving towards GA.  It's somewhat unfortunate that Money and OTel are both currently 0.9.x as adopting OTel's version would not indicate that there are breaking changes.  If we feel that this version alignment is worthwhile we can hold on making a proper release until early November when OTel 0.10.0 (or 1.0?) is released, in which case we can then align versions.  But we'd still have to consider Money patches and whether we'd increment the patch version apart from OTel.

Or we could keep the two decoupled, rev Money to 0.10.x and move forward.  If keeping up with OTel involves breaking changes we rev Money, otherwise we cut a patch release.

Either way, IMO it would be good to stabilize Money and to be able to cut a 1.0 release alongside a 1.0 release of OTel.

pauljamescleary: I think we go `0.10.0` and decouple from the OTel version, and I like the idea of doing a 1.0 release with OTel

HaloFour: Cool beans, updated this PR to `0.10.0-SNAPSHOT`.

ptravers: that makes sense to me

ptravers: @HaloFour do you want to release the updates for W3C Trace support before this release? I would probably prefer to get those out as a patch release before releasing this as they don't need to be lumped into this breaking release.

money-akka/src/main/scala/com/comcast/money/akka/SpanContextWithStack.scala
ptravers: I don't think the change would necessarily be correct. The end user could call these out of order and get the wrong span,

Push Span A get Scope A
Push Span B get Scope B
Call Scope A get Span B

HaloFour: Yes, that was already a problem if you called `push`/`pop` out of order.  It's still a problem in OTel if you call `push`/`close` out of order.  At most `Context` logs that you're reattaching the wrong context, but it allows it to happen.

I could follow a similar approach as `Context` and log (or throw?) if you attempt to close a scope out of order by comparing the top span before popping.  Does that alleviate your concern?

build.sbt
HaloFour: That's fine by me if you think that anyone would find it useful.

money-akka/src/main/scala/com/comcast/money/akka/SpanContextWithStack.scala
HaloFour: This is the comment from grpc Context if you attempt to pop the incorrect scope:

```java
    if (current() != toDetach) {
      // Log a severe message instead of throwing an exception as the context to attach is assumed
      // to be the correct one and the unbalanced state represents a coding mistake in a lower
      // layer in the stack that cannot be recovered from here.
      log.log(Level.SEVERE, "Context was not attached when detaching",
          new Throwable().fillInStackTrace());
    }
```

HaloFour: How does this approach feel?  It's similar to Context where it will log if the current state and expected state don't match indicating a stack imbalance.  If there is an unbalance it will still restore the previous state.

```scala
  override def push(span: Span): Scope = {
    // capture the previous stack of spans
    val previousStack = stack
    // push the new span to the head of the stack
    stack = span :: previousStack

    () => stack match {
      // happy path, the remaining stack of spans matches the captured stack of spans
      case _ :: rest if rest == previousStack =>
        stack = rest
      // unbalanced state, log the expected/actual head span and restore the stack to the captured stack of spans
      case head :: _ =>
        logSpanMismatch(span, head)
        stack = previousStack
      // unbalanced state, log the expected head span and leave the stack of spans empty
      case Nil =>
        logSpanMismatch(span, null)
    }
  }
```

ptravers (approved):

build.sbt
ptravers: I just thought that upstream legacy systems could be converted to a non money trace without re-instrumenting the software.

money-akka/src/main/scala/com/comcast/money/akka/SpanContextWithStack.scala
ptravers: Apologies for a slow reply. That works for me

build.sbt
HaloFour: I'll leave it to @pauljamescleary since he's the one with the keys to publish.

PR #144, opened 2020-10-12
Add code coverage configuration

ptravers:


HaloFour:
Shouldn't the change to this PR trigger the code coverage report to be posted after verify?  How can we confirm that this works?

pauljamescleary:
Hmm, what I had done for codecov was the following added on after the sbt build step...

```
      - name: Codecov
        uses: codecov/codecov-action@v1
        with:
          fail_ci_if_error: true # optional (default = false)
```

ptravers:
Clearly it doesn't work from here. I copied this from vinyldns since we are using the same github actions workflow

PR #148, opened 2020-10-19
Refactor SpanId API to force trace ID format to be consistent

HaloFour:
Refactors `SpanId` so that all trace IDs are generated or validated so that the rest of Money can safely make the assumption that all trace IDs are in UUIDv4 format.  Also ensures that span flags (such as sampling and state) are propagated to child spans.

Rather than using a constructor you use one of several static methods to obtain a `SpanId` that will either generate new IDs or reuse IDs from a parent `SpanId`:

`SpanId.createNew()` - creates a new trace with a new root span id
`SpanId.createChild(SpanId)` - creates a child span for the specified parent span
`SpanId.createRemote(String, long, long, byte, TraceState)` - creates a span id from remote/propagated ids, e.g. HTTP headers
`SpanId.fromSpanContext(SpanContext)` - gets a span id for the OpenTelemetry SpanContext
`SpanId.getInvalid` - gets an invalid span id


pauljamescleary (approved):
lgtm

PR #149, opened 2020-10-20
Breaks Formatters up until individual Formatter classes, propagates flags/state

HaloFour:
Breaks up the `Formatters` class into multiple `Formatter` implementations with a `FormatterChain` that, similarly to `HandlerChain`, can take a `Seq[Formatter]` and apply each of them to read/write HTTP headers for the given SpanID.   This could be a first pass at making the formatters configurable.

Reuses the OpenTelemetry `HttpTraceContext` implementation as a `Formatter` to read/write the `traceparent` and `tracestate` HTTP headers.


HaloFour:
Apologies for the rapid-fire PRs and the accumulation of commits into those PRs.  I've adopted this project as my Labweek project this week so I've been able to dedicate a lot of time towards it.  For this PR I can certainly rollback the commit that splits things out and makes it configurable and submit that as a separate PR.

I'm also on the fence as to whether `Formatter` should be moved to `money-api` and made into a more Java-friendly interface.

money-core/src/main/scala/com/comcast/money/core/formatters/FormatterChain.scala
```diff
@@ -0,0 +1,53 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.comcast.money.core.formatters
+import com.comcast.money.api.SpanId
+import com.typesafe.config.Config
+import scala.collection.JavaConverters._
+
+case class FormatterChain(formatters: Seq[Formatter]) extends Formatter {
```
ptravers, line 22:
```suggestion
final case class FormatterChain(formatters: Seq[Formatter]) extends Formatter {
```

money-core/src/main/scala/com/comcast/money/core/formatters/FormatterChain.scala
```diff
@@ -0,0 +1,53 @@
...
+    formatter => formatter.fromHttpHeaders(getHeader, log)
+  } headOption
+
+  override def fields: Seq[String] = formatters.flatMap {
+    formatter => formatter.fields
+  }
+
+  override def setResponseHeaders(getHeader: String => String, addHeader: (String, String) => Unit): Unit = formatters.foreach {
+    formatter => formatter.setResponseHeaders(getHeader, addHeader)
+  }
```
ptravers, line 37: do we want to offer some information as to the total size of headers we are adding?

money-core/src/main/scala/com/comcast/money/core/formatters/MoneyTraceFormatter.scala
```diff
@@ -0,0 +1,57 @@
...
+import io.opentelemetry.trace.{ TraceFlags, TraceState }
+
+import scala.util.{ Failure, Success, Try }
+
+object MoneyTraceFormatter {
+  private[core] val MoneyTraceHeader = "X-MoneyTrace"
+  private[core] val MoneyHeaderFormat = "trace-id=%s;parent-id=%s;span-id=%s"
+}
+
+class MoneyTraceFormatter extends Formatter {
```
ptravers, line 29:
```suggestion
final class MoneyTraceFormatter extends Formatter {
```

money-core/src/main/scala/com/comcast/money/core/formatters/OtelFormatter.scala
```diff
@@ -0,0 +1,68 @@
...
+    override def addEvent(name: String, attributes: Attributes): Unit = ()
+    override def addEvent(name: String, attributes: Attributes, timestamp: Long): Unit = ()
+    override def setStatus(canonicalCode: StatusCanonicalCode): Unit = ()
+    override def setStatus(canonicalCode: StatusCanonicalCode, description: String): Unit = ()
+    override def recordException(exception: Throwable): Unit = ()
+    override def recordException(exception: Throwable, additionalAttributes: Attributes): Unit = ()
+    override def updateName(name: String): Unit = ()
+    override def `end`(): Unit = ()
+    override def `end`(endOptions: EndSpanOptions): Unit = ()
+    override def isRecording: Boolean = false
```
ptravers, line 65: will this cause any issues in the propagator? we should probably create an integration test with otel and zipkin

ptravers (approved):

money-core/src/main/scala/com/comcast/money/core/formatters/FormatterChain.scala
HaloFour: The `fields` member would return the total list of headers that might be added.  The actual headers depend on the SpanId/SpanContext.

ptravers: I meant more that we might make the request breach max header size for lets say apache. wasnt sure if maybe it was worth giving the user some information about the size in bytes that we are adding

money-core/src/main/scala/com/comcast/money/core/formatters/OtelFormatter.scala
HaloFour: I don't believe so.  All of the OTel implementations grab the `SpanContext` and work with that.  There really isn't any useful information that the propagator could glean from the full `Span` API, the only member that returns a value is `isRecording`.

money-core/src/main/scala/com/comcast/money/core/formatters/FormatterChain.scala
HaloFour: I don't think that would be a concern.  Even if someone turned on every single propagator I've included I don't think the total size of the headers + values would exceed a few hundred bytes.

PR #151, opened 2020-10-22
Add TracerProvider and TracerProviderFactory implementations for OTel SPI

HaloFour:
Adds an implementation of `TracerProvider` and `TracerProviderFactory` that can be used with Java SPI service registration and automatically loaded by `OpenTelemetry.getTracer()`.

### DIFF
--- a/money-core/src/main/scala/com/comcast/money/core/MoneyTracerProvider.scala
+++ b/money-core/src/main/scala/com/comcast/money/core/MoneyTracerProvider.scala
@@ -19,7 +19,7 @@ package com.comcast.money.core
 import io.opentelemetry.trace
 import io.opentelemetry.trace.TracerProvider
 
-case class MoneyTracerProvider(tracer: Tracer) extends TracerProvider {
+final case class MoneyTracerProvider(tracer: Tracer) extends TracerProvider {
   override def get(instrumentationName: String): trace.Tracer = tracer
   override def get(instrumentationName: String, instrumentationVersion: String): trace.Tracer = tracer
 }
--- a/money-core/src/main/scala/com/comcast/money/core/MoneyTracerProvider.scala
+++ b/money-core/src/main/scala/com/comcast/money/core/MoneyTracerProvider.scala
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2012 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.comcast.money.core
+
+import io.opentelemetry.trace
+import io.opentelemetry.trace.TracerProvider
+
+case class MoneyTracerProvider(tracer: Tracer) extends TracerProvider {
+  override def get(instrumentationName: String): trace.Tracer = tracer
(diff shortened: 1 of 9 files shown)


HaloFour:
I'm still trying to see if this will integrate with opentelemetry-instrumentation.  So far it seems that the Java agent automatically configures the SDK reference implementation

money-core/src/main/scala/com/comcast/money/core/formatters/FormatterPropagator.scala
```diff
@@ -0,0 +1,48 @@
...
+import java.util
+
+import com.comcast.money.api.SpanId
+import io.grpc.Context
+import io.opentelemetry.context.propagation.TextMapPropagator
+import io.opentelemetry.trace.{ DefaultSpan, TracingContextUtils }
+
+import scala.collection.JavaConverters._
+
+case class FormatterPropagator(formatter: Formatter) extends TextMapPropagator {
```
ptravers, line 28:
```suggestion
final case class FormatterPropagator(formatter: Formatter) extends TextMapPropagator {
```

money-core/src/main/scala/com/comcast/money/core/formatters/FormatterPropagators.scala
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.comcast.money.core.formatters
+
+import io.opentelemetry.context.propagation.{ ContextPropagators, TextMapPropagator }
+
+case class FormatterPropagators(formatter: Formatter) extends ContextPropagators {
```
ptravers, line 21:
```suggestion
final case class FormatterPropagators(formatter: Formatter) extends ContextPropagators {
```

money-core/src/main/scala/com/comcast/money/core/MoneyTracerProvider.scala
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.comcast.money.core
+
+import io.opentelemetry.trace
+import io.opentelemetry.trace.TracerProvider
+
+case class MoneyTracerProvider(tracer: Tracer) extends TracerProvider {
```
ptravers, line 22:
```suggestion
final case class MoneyTracerProvider(tracer: Tracer) extends TracerProvider {
```

money-core/src/main/scala/com/comcast/money/core/MoneyTracerProvider.scala
```diff
@@ -0,0 +1,25 @@
...
+  override def get(instrumentationName: String): trace.Tracer = tracer
+  override def get(instrumentationName: String, instrumentationVersion: String): trace.Tracer = tracer
```
ptravers, line 24: do we want to support getting a specific version? I believe this should error out if there`s a version mismatch

ptravers (commented):
a few nits and actual q

money-core/src/main/scala/com/comcast/money/core/MoneyTracerProvider.scala
HaloFour: Let me check the docs on this interface to see exactly what is expected here.  Feels weird that if you're going through the trouble to set up SPI that you'd also make specific library+version requests to obtain a tracer.

HaloFour: From the JavaDocs it mentions that the name/version is for the instrumentation library, I don't think Money would care about that.

HaloFour: Looks like the SDK reference implementation returns a different `TracerSdk` instance for each requested instrumentation library.  It doesn't matter what the library names or versions are.  The library info is attached to the span and the exporter can inspect/use it, but that's all it does.

ptravers (approved):

PR #152, opened 2020-10-22
Refactor methods for generating/parsing/validating IDs to a helper class

HaloFour:
Pondering moving the logic for generating, validating and parsing of trace and span IDs into separate helper class so that `SpanId` can be less cluttered.  I don't know that I'm entirely happy with this but I thought I'd get some feedback.

### DIFF
--- a/money-api/src/main/java/com/comcast/money/api/IdGenerator.java
+++ b/money-api/src/main/java/com/comcast/money/api/IdGenerator.java
@@ -17,8 +17,8 @@
 package com.comcast.money.api;
 
 import java.util.Locale;
-import java.util.Random;
 import java.util.UUID;
+import java.util.concurrent.ThreadLocalRandom;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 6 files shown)


money-api/src/main/java/com/comcast/money/api/IdGenerator.java
```diff
@@ -0,0 +1,140 @@
...
+
+    /**
+     * Generates a random trace ID.
+     */
+    public static String generateRandomTraceId() {
+        long hi, lo;
+        Random random = threadLocalRandom.get();
+        do {
+            hi = random.nextLong();
+            lo = random.nextLong();
```
ptravers, line 47: do we know how random this is?

ptravers (commented):
looks good. just one q around the safety of nextLong

money-api/src/main/java/com/comcast/money/api/IdGenerator.java
HaloFour: Does it help that this is the implementation that the OTel SDK uses?

https://github.com/open-telemetry/opentelemetry-java/blob/master/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/RandomIdsGenerator.java#L31

I did notice that `ThreadLocalRandom` is actually a class provided by Java 8, I can at least switch the impl to using that.

ptravers: Yeah I would be okay with defaulting to what otel does here. I won't have time to dig in to this tonight. But I have set aside some time early tomorrow morning to double check

ptravers (commented):
Will double check the randomness safety

ptravers:
I like the refactor as it exposes some critical logic

ptravers:
this is likely preferable to minimize contention and garbage collection

http://thoughtfuljava.blogspot.com/2012/09/prefer-threadlocalrandom-over-random.html

ptravers (approved):

PR #153, opened 2020-10-23
WIP: Tracer per instrumentation library, flow through to SpanInfo

HaloFour:
Explore handling `TracerProvider` and instrumentation library info "correctly".  It involves being able to provide a different tracer per library name+version and each span created by that tracer will have that library info written into it.  That will then be copied into the SpanInfo.

### DIFF
--- a/money-api/src/main/java/com/comcast/money/api/InstrumentationLibrary.java
+++ b/money-api/src/main/java/com/comcast/money/api/InstrumentationLibrary.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2012 Comcast Cable Communications Management, LLC
+ *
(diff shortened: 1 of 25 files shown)


HaloFour:
It might be possible to simplify by exposing the instrumentation library on the span interface and making it mutable, then I might be able to use a decorator to update the library info on the span after the span is created vs. creating new tracers.  But this is as far as I've gotten tonight.

ptravers:
@HaloFour  do you still need me to look at this? after the last PR

HaloFour:
Obsolete

HaloFour:
Nope, the other PR took care of the same concerns in a cleaner way.

PR #154, opened 2020-10-23
Wire instrumentation library info through SpanFactory to SpanInfo

HaloFour:
To support SPI scenarios where an instrumentation library will request a tracer with the instrumentation library information which may be used by the span handler/exporter wire the instrumentation library information (name and version) through the CoreSpan and SpanInfo.

### DIFF
--- a/money-otlp-exporter/src/test/java/com/comcast/money/otel/handlers/otlp/OtlpHandlerSpec.java
+++ b/money-otlp-exporter/src/test/java/com/comcast/money/otel/handlers/otlp/OtlpHandlerSpec.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2012 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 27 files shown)


HaloFour:
Taking WIP off of this PR, I think this approach works and has relatively limited impact on the APIs.

TL;DR:  The instrumentation library is exposed from `SpanInfo` but there is nothing in the API that specifies how it gets there.  The `CoreSpanFactory` defaults it to the name/version of Money, but `MoneyTraceProvider` can create new instances of `Tracer` and copy the `CoreSpanFactory` from the existing singleton and apply different instrumentation library.

money-otlp-exporter/src/test/java/com/comcast/money/otel/handlers/otlp/OtlpHandlerSpec.java
```diff
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2012 Comcast Cable Communications Management, LLC
```
ptravers, line 2: this should really be 2020

ptravers (approved):
not really related but we need to bump the year in the headers

PR #155, opened 2020-10-23
Add configurable span sampling

HaloFour:
Adds a basic sampler implementation with a configurable sampler.  The sampler implementation decides if a span should be sampled, recorded or dropped.  A dropped span is not recorded, does not invoke the SpanHandler and doesn't retain any state except for the SpanId for propagation.  Recorded and sampled spans both retain all state (notes, etc.) and invoke the SpanHandler, the difference is the tracing flags on the SpanId which indicates that the span is sampled which is propagated in headers to upstream services and can be interpreted by the SpanHandler.


HaloFour:
Needs unit tests but I'd love feedback on the direction.  The one thing this implementation doesn't do that the OTel implementation does is allow the sampler to record notes onto the span.  That is used by the probability sampler to denote the probability percentage.

money-core/src/main/scala/com/comcast/money/core/samplers/ParentBasedSampler.scala
```diff
@@ -0,0 +1,41 @@
+package com.comcast.money.core.samplers
+import com.comcast.money.api.SpanId
+import com.typesafe.config.Config
+
+/**
+ * A sampler that uses the parent span sampling decision if one exists, otherwise uses the root sampler
+ * to determine the sampler result.
+ */
+class ParentBasedSampler extends ConfigurableSampler {
```
ptravers, line 9:
```suggestion
final class ParentBasedSampler extends ConfigurableSampler {
```

money-core/src/main/scala/com/comcast/money/core/samplers/RatioBasedSampler.scala
```diff
@@ -0,0 +1,41 @@
...
+
+package com.comcast.money.core.samplers
+
+import com.comcast.money.api.{ Note, SpanId }
+
+/**
+ * A sampler that calculates that a ratio of spans should be recorded based on the lower 64-bits of the trace id.
+ * @param ratio of spans to be recorded and sampled, between 0.0 and 1.0
+ */
+class RatioBasedSampler(val ratio: Double) extends Sampler {
```
ptravers:
```suggestion
final class RatioBasedSampler(val ratio: Double) extends Sampler {
```

money-core/src/main/scala/com/comcast/money/core/samplers/SamplerResult.scala
```diff
@@ -0,0 +1,48 @@
...
+ * context but the span will not be reported to the [[com.comcast.money.api.SpanHandler]] and no notes will be recorded.
+ */
+case object DropResult extends SamplerResult
+
+/**
+ * Specifies that the span will be recorded along with all notes.
+ * @param sample indicates whether the span will be marked as sampled which will be propagated to upstream systems
+ * @param notes to be recorded on the span
+ */
+sealed case class RecordResult(sample: Boolean = true, notes: Seq[Note[_]] = Nil) extends SamplerResult
```
ptravers, line 37: Why sealed?

money-core/src/main/scala/com/comcast/money/core/samplers/ParentBasedSampler.scala
```diff
@@ -0,0 +1,57 @@
...
+    sampler.shouldSample(spanId, parentSpanId, spanName)
+  }
+
+  override def configure(conf: Config): Unit = {
+    root = SamplerFactory.create(conf.getConfig("root"))
+    remoteSampled = findSampler(conf, "remote-sampled", remoteSampled)
+    remoteNotSampled = findSampler(conf, "remote-not-sampled", remoteNotSampled)
+    localSampled = findSampler(conf, "local-sampled", localSampled)
+    localNotSampled = findSampler(conf, "local-not-sampled", localNotSampled)
+  }
```
ptravers, line 51: When would we want to update the config for this class?

ptravers (commented):
Looks good! a few qs

money-core/src/main/scala/com/comcast/money/core/samplers/SamplerResult.scala
HaloFour: I think I had planned on having another hierarchy of case objects below `RecordResult` but I never got around to it.

money-core/src/main/scala/com/comcast/money/core/samplers/ParentBasedSampler.scala
HaloFour: The OTel version of this sampler is configurable in the same manner, to allow overriding the parent span sampling flag.

money-core/src/main/scala/com/comcast/money/core/samplers/SamplerResult.scala
HaloFour: Except that given case hierarchy is illegal I guess that wouldn't be possible anyway. 😁

ptravers: Ha yeah you don't ever want to inherit from a case class

money-core/src/test/scala/com/comcast/money/core/samplers/RatioBasedSamplerSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+    "records the sampling probability as a note" in {
+      val spanId = createSpanIdWithLoBits(0L)
+
+      val underTest = new RatioBasedSampler(1.0)
+      val result = underTest.shouldSample(spanId, None, "name")
+
+      result should matchPattern {
+        case RecordResult(true, Seq(note)) if note.name == "sampling.probability" && note.value == 1.0 =>
+      }
+    }
```
ptravers, line 68: neat, I hadn't seen this before

ptravers (approved):

PR #156, opened 2020-10-24
Update to OpenTelemetry 0.10.0

HaloFour:
Updates to using OpenTelemetry 0.10.0 libraries.

Summary of major changes:

1. Packages renamed from `io.opentelemetry.trace._` to `io.opentelemetry.api.trace._`
2. `StatusCanonicalCode` renamed to `StatusCode`
3. Many methods on `Span` to add attributes, events, etc., changed to return the instance of the `Span` instead of `void`/`Unit`
    * Because of this I updated the Money `Span` interface to override all of these members and use covariant returns so that they return a Money `Span` instead of the OpenTelemetry `Span`.
4. `TextMapPropagator.Getter[T]` now has two required members and is no longer valid as a functional interface in Java or as a SAM in Scala.
    * The second member is `keys()` which returns an `Iterable[String]` of header names.  This forces the `Formatter#fromHttpHeaders` method to also be changed to accept an `Iterable[String]` of the header names.
5. `io.grpc.Context` has been replaced with `io.opentelemetry.context.Context`
6. Context APIs have been removed, e.g. `Tracer#withSpan`, `TracingContextUtils`
7. Context APIs have been moved to `Context`, e.g. `Context.with(span).makeCurrent()`
    * Because of this there is no longer an API path to intercept when the current span is changed, so an SPI provider for `ContextStorage` is used to replace the underlying storage mechanism which can intercept when contexts are changed which detects the span in those contexts and updates MDC accordingly.


HaloFour:
The biggest change is to how span thread context is managed.  The GRPC Context has been replaced with a built-in implementation and the static helper class has been replaced with a combination of static and instance methods.  The trickiest aspect is that putting a span into the context involves calling an instance method on `Span`, which doesn't really work when `Span` is mocked.

The rest is mostly superficial.

ptravers:
Ooof thanks for doing this!

HaloFour:
Of course they decided to change the packages for everything, so the PR looks huge now.  😄

HaloFour:
I'm taking the WIP label off.  I've adopted the late-breaking changes and switched the dependencies to the release build (and removing the snapshot repository).

ptravers (approved):

PR #157, opened 2020-10-25
Update to Mockito 3.4

HaloFour:
Updates tests to using Mockito 3.4.  This is good for tech debt in general but it's also in anticipation of updating to OpenTelemetry 0.10.0 which is looking at putting core functionality behind default method implementations in interfaces which Mockito 1.x does not support calling through to when that interface is mocked.

I had to replace Springockito with SpringBootTest for mocking beans with the aspect in the Spring tests and I had to switch from `doReturn(...).when(...)` to `when(...).thenReturn(...)` due to how Scala resolves overload candidates with varargs, but otherwise there were few required changes.

### DIFF
--- a/money-spring/src/test/java/com/comcast/money/spring/TracedMethodInterceptorSpec.java
+++ b/money-spring/src/test/java/com/comcast/money/spring/TracedMethodInterceptorSpec.java
@@ -63,9 +63,6 @@ public class TracedMethodInterceptorSpec {
 
     @Before
     public void setUp() {
-        // Needed to init the Argument Captor
-        // MockitoAnnotations.initMocks(this);
-
         when(springTracer.spanBuilder(anyString())).thenReturn(spanBuilder);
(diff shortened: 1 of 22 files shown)


money-spring/src/test/java/com/comcast/money/spring/TracedMethodInterceptorSpec.java
```diff
@@ -55,13 +61,10 @@
...
     private Scope scope;
 
-    @Captor
-    private ArgumentCaptor<Boolean> spanResultCaptor;
-
     @Before
     public void setUp() {
         // Needed to init the Argument Captor
-        MockitoAnnotations.initMocks(this);
+        // MockitoAnnotations.initMocks(this);
```
ptravers, line 67: I dont think you need this

ptravers (changes requested):
one small change! thanks so much for this

ptravers (approved):

PR #158, opened 2020-10-26
Add X-MoneySampled header to MoneyTraceFormatter to convey sampled state

HaloFour:
Adds an optional `X-MoneySampled` headers with a value of `0` or `1` to convey whether the remote span is sampled.

Another option would be to add another key-value pair to the end of the existing header but that will break the Go Money implementation which assumes that there can only be three key-value pairs.

### DIFF
--- a/money-core/src/main/scala/com/comcast/money/core/formatters/MoneyTraceFormatter.scala
+++ b/money-core/src/main/scala/com/comcast/money/core/formatters/MoneyTraceFormatter.scala
@@ -17,41 +17,52 @@
 package com.comcast.money.core.formatters
 
 import com.comcast.money.api.SpanId
-import io.opentelemetry.trace.{ TraceFlags, TraceState }
+import io.opentelemetry.trace.{TraceFlags, TraceState}
 
-import scala.util.{ Failure, Success, Try }
+import scala.util.{Failure, Success, Try}
 
 object MoneyTraceFormatter {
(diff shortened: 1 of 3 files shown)


HaloFour:
I've opened this issue with `golang-money` to inquire about modifying the existing `X-MoneyTrace` header instead of adding a second header:

https://github.com/xmidt-org/golang-money/issues/43

HaloFour:
TL;DR:

Both the Scala and Go versions of Money expect a header with the name `X-MoneyTrace`.  The header contains a series of name/value pairs delimited by `;` where the key and value is split by a single `=`.

The Scala version requires a minimum of three key/value pairs, but only inspects the first three pairs.  It does not inspect the names of the keys at all, rather it assumes that the pairs will always be in the order of `trace-id`, `parent-id` and `span-id`.

Conversely, the Go version requires exactly three key/value pairs.  However, does inspect the names of the keys to determine which is the `trace-id`, `parent-id` and `span-id`.  It must see each of the keys exactly once.  Because of these rules it cannot accept additional key/value pairs.

The only way to keep the Scala and Go versions of Money compatible with each other and previous versions would be to leave the `X-MoneyTrace` header alone.  It will always emit three key/value pairs and they will always be in the order `trace-id`, `parent-id` and `span-id`.

HaloFour:
I'm going to close this PR in favor of https://github.com/Comcast/money/pull/159

PR #159, opened 2020-10-27
WIP: Add X-MoneyState header for sampling and state propagation

HaloFour:
Adds an `X-MoneyState` header for propagating additional span information.  This header will include a series of name/value pairs.  Recognized entries can be interpreted directly and the remainder will be appended to the `TraceState` of the span.

This will enable adding a sampling flag without affecting the entries or format of the `X-MoneyTrace` flag which may break one or more of the alternate implementations of Money.

### DIFF
--- a/money-core/src/main/scala/com/comcast/money/core/formatters/MoneyTraceFormatter.scala
+++ b/money-core/src/main/scala/com/comcast/money/core/formatters/MoneyTraceFormatter.scala
@@ -17,11 +17,11 @@
 package com.comcast.money.core.formatters
 
 import com.comcast.money.api.SpanId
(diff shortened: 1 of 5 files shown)


ptravers:
I don't totally see the need to add another header, why not just use the w3c header if you are changing header?

HaloFour:
There is a separate formatter for using the W3C Trace Context header.

HaloFour:
Maybe there isn't much of a need to update the header.

The issue would be that if Money continues to prioritize the `X-MoneyTrace` header over other headers (as it did when the ZipKin B3 headers were added) that would mean that any sampling information would be lost and the new span would have to be sampled by default.  This isn't the end of the world, it's what Money does today.

The formatter changes and configuration does enable you to prioritize the formatters by specifying their order.  Whichever formatter can successfully interpret a span from the headers gets used first.  As of right now the Money header is first by default, but we could make the W3C Trace Context formatter first instead.

ptravers:
Yeah I think that makes sense are you comfortable with closing this PR given that?

HaloFour:
Considering other options

PR #160, opened 2020-10-27
Adds configurable formatters and ingress/egress-specific formatters

HaloFour:
Adds support for a Formatter to be configurable.  Also adds "ingress" and "egress" formatter decorators which are a Formatter implementation which only reads or writes HTTP headers for trace propagation respectively.

The following would be a configuration that would cause Money to only read W3C Trace Context headers from downstream requests, but that will write W3C Trace Context headers and Money Trace headers to upstream services.

```
formatting = {
    formatters = [
        {
            type = ingress
            formatters = [
                { type = trace-context }
            ]
        },
        {
            type = egress
            formatters = [
                { type = trace-context },
                { type = money-trace }
            ]
        },
    ]
}
```

### DIFF
--- a/money-core/src/main/scala/com/comcast/money/core/formatters/EgressFormatter.scala
+++ b/money-core/src/main/scala/com/comcast/money/core/formatters/EgressFormatter.scala
@@ -1,0 +1,12 @@
+package com.comcast.money.core.formatters
+import com.comcast.money.api.SpanId
+
+/**
+ * Decorates a formatter so that it writes headers for outgoing requests but does not interpret headers for incoming requests.
+ */
+class EgressFormatter(val formatter: Formatter) extends Formatter {
+  override def toHttpHeaders(spanId: SpanId, addHeader: (String, String) => Unit): Unit =
+    formatter.toHttpHeaders(spanId, addHeader)
+  override def fromHttpHeaders(getHeader: String => String, log: String => Unit): Option[SpanId] = None
+  override def fields: Seq[String] = Nil
+}
--- a/money-core/src/main/scala/com/comcast/money/core/formatters/EgressFormatter.scala
+++ b/money-core/src/main/scala/com/comcast/money/core/formatters/EgressFormatter.scala
@@ -20,7 +20,7 @@ import com.comcast.money.api.SpanId
 /**
  * Decorates a formatter so that it writes headers for outgoing requests but does not interpret headers for incoming requests.
  */
-class EgressFormatter(val formatter: Formatter) extends Formatter {
+final class EgressFormatter(val formatter: Formatter) extends Formatter {
   override def toHttpHeaders(spanId: SpanId, addHeader: (String, String) => Unit): Unit =
     formatter.toHttpHeaders(spanId, addHeader)
   override def fromHttpHeaders(getHeader: String => String, log: String => Unit): Option[SpanId] = None
--- a/money-core/src/main/scala/com/comcast/money/core/formatters/EgressFormatter.scala
+++ b/money-core/src/main/scala/com/comcast/money/core/formatters/EgressFormatter.scala
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2012 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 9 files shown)


money-core/src/main/scala/com/comcast/money/core/formatters/EgressFormatter.scala
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+
+package com.comcast.money.core.formatters
+import com.comcast.money.api.SpanId
+
+/**
+ * Decorates a formatter so that it writes headers for outgoing requests but does not interpret headers for incoming requests.
+ */
+class EgressFormatter(val formatter: Formatter) extends Formatter {
```
ptravers, line 23:
```suggestion
final class EgressFormatter(val formatter: Formatter) extends Formatter {
```

money-core/src/main/scala/com/comcast/money/core/formatters/IngressFormatter.scala
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package com.comcast.money.core.formatters
+import com.comcast.money.api.SpanId
+
+/**
+ * Decorates a formatter so that it interprets headers for incoming requests but does not write headers for outgoing requests.
+ */
+final class IngressFormatter(val formatter: Formatter) extends Formatter {
+  override def toHttpHeaders(spanId: SpanId, addHeader: (String, String) => Unit): Unit = ()
```
ptravers, line 24: does this not need to be implemented or am I missing something:?

ptravers, line 24: I think I would rather fail loudly if someone calls this than do nothing silently

money-core/src/main/scala/com/comcast/money/core/formatters/EgressFormatter.scala
```diff
@@ -0,0 +1,28 @@
...
+package com.comcast.money.core.formatters
+import com.comcast.money.api.SpanId
+
+/**
+ * Decorates a formatter so that it writes headers for outgoing requests but does not interpret headers for incoming requests.
+ */
+final class EgressFormatter(val formatter: Formatter) extends Formatter {
+  override def toHttpHeaders(spanId: SpanId, addHeader: (String, String) => Unit): Unit =
+    formatter.toHttpHeaders(spanId, addHeader)
+  override def fromHttpHeaders(getHeader: String => String, log: String => Unit): Option[SpanId] = None
```
ptravers, line 26: I think I would rather fail loudly if someone calls this than do nothing silently

ptravers (commented):
What do you think the benefits are of doing nothing in the event someone calls an unsupported method in the formatters??

ptravers (commented):
How would you feel about adding traits for Ingress and Egress formatting that have the to and from methods on them respectively? Only concern would be blast radius of the change. But I would prefer compile time failure versus runtime no op

money-core/src/main/scala/com/comcast/money/core/formatters/IngressFormatter.scala
HaloFour: They're supposed to be called, in that Formatters assume that they both read and write headers.  But for these two Formatters they're special in that they either read or write headers, not both.   But the Formatters they wrap don't know this.

ptravers: I get that, I guess my concern would be a user calls the wrong method and gets not feedback telling them that it's wrong

HaloFour: A user isn't going to call them directly, instrumentation is and it won't know that the formatter isn't supposed to read or write certain headers, especially if these formatters are used as a part of a FormatterChain.

ptravers: gotcha!

ptravers: can we make them private to money?

ptravers (approved):

PR #161, opened 2020-10-29
Refactor CoreSpanFactory to call through CoreSpanBuilder

HaloFour:
Refactors the relationship between CoreSpanFactory and CoreSpanBuilder so that CoreSpanFactory uses CoreSpanBuilder to actually create the spans instead of vice versa.  The reason for this is that CoreSpanBuilder offers a number of members to mutate the span before it is created, after which those members should be readonly.  If CoreSpanBuilder uses a SpanFactory to create the Span instances that requires that the Span members be mutable so that the CoreSpanBuilder can update them on the created Span.

This is a prelude to implementing a few other features of OTel such as adding Links to a Span through the Span.Builder.

### DIFF
--- a/money-api/src/main/java/com/comcast/money/api/Span.java
+++ b/money-api/src/main/java/com/comcast/money/api/Span.java
@@ -73,11 +73,6 @@ public interface Span extends io.opentelemetry.trace.Span, Scope {
     void stopTimer(String timerKey);
 
     /**
-     * Updates the kind of the span
-     */
-    void updateKind(io.opentelemetry.trace.Span.Kind kind);
-
-    /**
      * Attaches a {@link Scope} to the span which will be closed when the span is stopped
      */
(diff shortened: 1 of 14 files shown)


HaloFour:
I'm also tempted to remove the `start` method from the `Span` interface.  The `Tracer.startSpan` API returns an already started Span, as does the OTel `Span.Builder` API.

HaloFour:
Went ahead and removed that `Span#start` method.  The only way you could get an unstarted `Span` from Money previously was to use the `SpanFactory` directly.  Any `Span` exposed from Money would be already started.  This is also the case with OTel, so removal of this method makes them act more consistently with one another.

ptravers:
Is there a reason to expose the SpanFactory at all? what's the use case for that?

ptravers (approved):

PR #168, opened 2020-12-21
Adds context filter to make MDC support pluggable

HaloFour:
Adds OpenTelemetry ContextStorage filter to make MDC support pluggable and allow any plugin to intercept when the context (and Span) changes for the current thread.  This would enable applications to write their own MDC plugins or store other information in the thread context.

Also refactors the factory pattern used with SpanHandler, Formatter, Sampler and now ContextStorageFilter to a single helper trait that can find and instantiate plugins from a standard configuration format.  This trait also looks for factory methods or constructors that accept a Config so that the plugin can configure itself on start rather than a separate method and mutable state.


money-core/src/main/scala/com/comcast/money/core/ConfigurableTypeFactory.scala
```diff
@@ -0,0 +1,124 @@
...
+    } map { constructor => _: Config => constructor.newInstance().asInstanceOf[T] }
+
+  private def findStaticInstance(cls: Class[_]): Option[Config => T] =
+    cls.getFields.find {
+      field =>
+        Modifier.isStatic(field.getModifiers) &&
+          field.getType == cls &&
+          field.getName == MODULE_FIELD_NAME
+    } map { field => _: Config => field.get(null).asInstanceOf[T] }
+}
```
ptravers, line 124: How much code does this save us? I feel like the abstraction itself is deeper than what is being abstracted. Or at least that is my first impression. would you agree with that at all?

ptravers: If we want to go this route we might be better served with a macro. but you would need to bear in mind that that api is being fully rewritten for scala 3.

ptravers (commented):
Took a first pass. Not in love with the usage of reflection particularly in this context with what I believe to be a not insubstantial level of performance sensitivity.

ptravers:
Still looking through though. I apologise for my slowness;

money-otel-logging-exporter/src/main/scala/com/comcast/money/otel/handlers/logging/LoggingSpanHandler.scala
```diff
@@ -31,27 +31,19 @@ import io.opentelemetry.sdk.trace.`export`.SpanExporter
  *     async = true
  *     handlers = [
  *       {
- *         class = "com.comcast.money.otel.handlers.jaeger.JaegerOtelSpanHandler"
+ *         class = "com.comcast.money.otel.handlers.logging.LoggingSpanHandler"
```
ptravers, line 34: :+1:

money-core/src/main/scala/com/comcast/money/core/ConfigurableTypeFactory.scala
```diff
@@ -0,0 +1,124 @@
...
+
+  private def findFactoryMethod(cls: Class[_]): Option[Config => T] =
+    cls.getMethods.find {
+      method =>
+        Modifier.isStatic(method.getModifiers) &&
+          method.getParameterCount == 1 &&
+          method.getParameterTypes()(0) == classOf[Config] &&
+          method.getReturnType == cls &&
+          method.getName == APPLY_NAME
+    } map { method => config: Config => method.invoke(null, config).asInstanceOf[T] }
```
ptravers, line 103: infix with alphabetic methods is deprecated in scala 3
```suggestion
    }
    .map { method => config: Config => method.invoke(null, config).asInstanceOf[T] }
```

money-core/src/main/scala/com/comcast/money/core/ConfigurableTypeFactory.scala
```diff
@@ -0,0 +1,124 @@
...
+
+  private def findConfigConstructor(cls: Class[_]): Option[Config => T] =
+    cls.getConstructors.find {
+      constructor =>
+        constructor.getParameterCount == 1 &&
+          constructor.getParameterTypes()(0) == classOf[Config]
+    } map { constructor => config: Config => constructor.newInstance(config).asInstanceOf[T] }
```
ptravers, line 110:
```suggestion
    }
    .map { constructor => config: Config => constructor.newInstance(config).asInstanceOf[T] }
```

money-core/src/main/scala/com/comcast/money/core/ConfigurableTypeFactory.scala
```diff
@@ -0,0 +1,124 @@
...
+
+  private def findDefaultConstructor(cls: Class[_]): Option[Config => T] =
+    cls.getConstructors.find {
+      _.getParameterCount == 0
+    } map { constructor => _: Config => constructor.newInstance().asInstanceOf[T] }
```
ptravers, line 115:
```suggestion
    }
    .map { constructor => _: Config => constructor.newInstance().asInstanceOf[T] }
```

money-core/src/main/scala/com/comcast/money/core/ConfigurableTypeFactory.scala
```diff
@@ -0,0 +1,124 @@
...
+
+  private def findStaticInstance(cls: Class[_]): Option[Config => T] =
+    cls.getFields.find {
+      field =>
+        Modifier.isStatic(field.getModifiers) &&
+          field.getType == cls &&
+          field.getName == MODULE_FIELD_NAME
+    } map { field => _: Config => field.get(null).asInstanceOf[T] }
```
ptravers, line 123:
```suggestion
    }
    .map { field => _: Config => field.get(null).asInstanceOf[T] }
```

HaloFour:
> Not in love with the usage of reflection particularly in this context with what I believe to be a not insubstantial level of performance sensitivity.

The factories and reflection code would only apply during bootstrap.  Once that's done the plugin instances would be initialized and configured and used directly.

money-core/src/main/scala/com/comcast/money/core/ConfigurableTypeFactory.scala
HaloFour: It's true, it's a bit more code than what it replaced, but it does two things.

1.  It replaces the separate uses of reflection to manually create instances of a class by name used in the different factory objects so that they all work the same way.
2. It eliminates the separate configuration step (and configuration traits) used to provide configuration to the plugin instance.  Now configuration has to happen as a part of initialization making it impossible to end up with a plugin instance that isn't configured.  This enables plugins to be properly immutable.

ptravers (approved):

PR #170, opened 2021-01-21
Make MDC context storage configurable, add unit tests

HaloFour:
Adds formatted and structured MDC context storage filters with configuration.

### DIFF
--- a/money-core/src/main/scala/com/comcast/money/core/context/StructuredMdcContextStorageFilter.scala
+++ b/money-core/src/main/scala/com/comcast/money/core/context/StructuredMdcContextStorageFilter.scala
@@ -17,46 +17,49 @@
 package com.comcast.money.core.context
 
 import com.comcast.money.api.SpanInfo
-import com.comcast.money.core.context.StructuredMdcContextStorageFilter.{ DefaultParentIdKey, DefaultSpanIdKey, DefaultSpanNameKey, DefaultTraceIdKey }
 import com.comcast.money.core.internal.{ SpanContext, SpanLocal }
-import com.typesafe.config.Config
+import com.typesafe.config.{ Config, ConfigFactory }
 import org.slf4j.MDC
 import org.slf4j.spi.MDCAdapter
 
 object StructuredMdcContextStorageFilter {
-  private val DefaultTraceIdKey = "trace-id"
-  private val DefaultSpanIdKey = "span-id"
-  private val DefaultParentIdKey = "parent-id"
-  private val DefaultSpanNameKey = "span-name"
+  private val DefaultConfig = ConfigFactory.parseString(
+    """
+      | format-ids-as-hex = false
+      | trace-id = "trace-id"
+      | span-id = "span-id"
+      | parent-id = "parent-id"
+      | span-name = "span-name"
+      |""".stripMargin)
+
+  private val TraceIdKey = "trace-id"
+  private val SpanIdKey = "span-id"
+  private val ParentIdKey = "parent-id"
+  private val SpanNameKey = "span-name"
   private val FormatIdsAsHexKey = "format-ids-as-hex"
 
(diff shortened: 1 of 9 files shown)


money-core/src/test/scala/com/comcast/money/core/context/ContextStorageFilterFactorySpec.scala
```diff
@@ -0,0 +1,79 @@
...
+import org.scalatestplus.mockito.MockitoSugar
+
+class ContextStorageFilterFactorySpec extends AnyWordSpec with Matchers with MockitoSugar {
+
+  "ContextStorageFilterFactory" should {
+    "return an instance of mdc filter" in {
+      val config = ConfigFactory.parseString("type = \"mdc\"")
+
+      val filter = ContextStorageFilterFactory.create(config)
+      inside(filter) {
```
ptravers, line 33: I hadn't seen this before. nice!

money-core/src/main/scala/com/comcast/money/core/context/StructuredMdcContextStorageFilter.scala
```diff
@@ -0,0 +1,81 @@
...
+    val formatIdsAsHex = conf.hasPath(FormatIdsAsHexKey) &&
+      conf.getBoolean(FormatIdsAsHexKey)
+
+    new StructuredMdcContextStorageFilter(spanContext, mdc, traceIdKey, spanIdKey, parentIdKey, spanNameKey, formatIdsAsHex)
+  }
+
+  private def readConfString(conf: Config, key: String, defaultValue: String) =
+    if (conf.hasPath(key)) {
+      conf.getString(key)
+    } else defaultValue
```
ptravers, line 50: I think since this happens at boot of the application. we should throw rather than have a default value.

ptravers (commented):
My only comment is that having a default in the case of improper configuration feels smelly to me and not obvious to the user that their configuration doesn't do what they intended

money-core/src/main/scala/com/comcast/money/core/context/StructuredMdcContextStorageFilter.scala
HaloFour: I'm treating the config more like overrides here, so if you don't specify the value in the config it just uses the default.

HaloFour:
I've switched to using a fallback config to establish the default values rather than checking to see if a key exists at the specific path.  This seems to be the more recommended approach according to the [readme](https://github.com/lightbend/config#how-to-handle-defaults).

money-core/src/test/scala/com/comcast/money/core/context/ContextStorageFilterFactorySpec.scala
```diff
@@ -0,0 +1,79 @@
...
+           |""".stripMargin)
+
+      val formatter = ContextStorageFilterFactory.create(config)
+      inside(formatter) {
+        case Some(f: ConfiguredContextStorageFilter) =>
+          f.config shouldBe config
+      }
+    }
+
+    "returns a disabled filter on an unknown type" in {
```
ptravers, line 72: Yeah, I still don't like this behaviour. I think it's not obvious to the user that this has happened. I would prefer to error out if a filter is not known

ptravers (changes requested):
I think the behaviour of silently failing for unknown filters could result in hard to diagnose issues for our users

money-core/src/test/scala/com/comcast/money/core/context/ContextStorageFilterFactorySpec.scala
HaloFour: That's probably fair.  All of the factories bail in the same way due to bad config, I could update them all to throw/fail instead.

ptravers: Thanks so much for doing that with https://github.com/Comcast/money/pull/171

ptravers (approved):

PR #175, opened 2021-02-14
Remove Scala dependency in money-api

HaloFour:
The first (and definitely easiest) of what will likely be a long line of painful PRs.  Removes the Scala dependency from the money-api project.

Breaking change:  `SpanBuilder#setParent(Option<Span>)` changed to `SpanBuilder#setParent(Optional<Span>)`, although I might want to just remove this API in favor of the overload that accepts `Context`, as OpenTelemetry has.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -12,6 +12,8 @@ lazy val copyApiDocsTask = taskKey[Unit]("Copies the scala docs from each projec
 
 lazy val props = new SystemProperties()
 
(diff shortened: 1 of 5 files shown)


HaloFour:
sbt-coverage plugin requires a release to support Scala 2.12.3:

https://github.com/scoverage/sbt-scoverage/issues/319

ptravers (approved):

PR #181, opened 2021-06-29
Bump to safe versions

ptravers:
Fixes dependabot alerts :crossed_fingers: 

Not critical since it's the sample code but not a great sign as to the recency of our sample code...


HaloFour (approved):
👍

PR #182, opened 2021-06-30
Bump patch releases of scala

ptravers:


HaloFour:
I believe you'll need to bump the scala versions in the GHA matrix as well

ptravers:
Good shout!

